Issue #12, opened 2016-08-14
Fix type spec hacks


yurrriq:
Related #5, #8.

- [x] [`prometheus_metric:value()`](https://github.com/deadtrickster/prometheus.erl/blob/60bc05871388d9fbed9c76b88e4178d7f0812cb1/src/prometheus_metric.erl#L39-L41)
- [x] Introduce separate types for [`prometheus_collector:callback()`](https://github.com/deadtrickster/prometheus.erl/blob/60bc05871388d9fbed9c76b88e4178d7f0812cb1/src/prometheus_collector.erl#L15-L18)
  - [x] `mf_callback()`
  - [x] `registry_callback()`
- [x] [`prometheus_collector:collect_mf/2`](https://github.com/deadtrickster/prometheus.erl/blob/60bc05871388d9fbed9c76b88e4178d7f0812cb1/src/prometheus_collector.erl#L20-L21) (namely the return type)  
    - There's no quick fix due to `prometheus_vm_memory_collector:collect_mf/2`.  
    - #17 removes the `FIXME`
- [x] [`prometheus_collector:collect_metrics/2`](https://github.com/deadtrickster/prometheus.erl/blob/60bc05871388d9fbed9c76b88e4178d7f0812cb1/src/prometheus_collector.erl#L23-L27)
  - [x] Add real type for `Data`
    - addressed in https://github.com/deadtrickster/prometheus.erl/commit/175eba40fbf7761e8166bd7c361a22ac3d16120d
  - [x] Replace `#'Metric'{}` with `prometheus_model:'Metric'()`
    - Addressed in #16
  - [x] Decide whether `Metrics` should really be one or many https://github.com/deadtrickster/prometheus.erl/commit/4b0de539ca2da872298bab92c940e7945bb15a0f
- [x] Refine `Labels` types in `prometheus_model_helpers`. See `FIXME` tags.

deadtrickster:
maybe introduce separate callback types for `prometheus_collector:callback`? like `mf_callback` and `registry_callback`?

yurrriq:
That sounds good to me. The more specific we can get with the types the better.

yurrriq:
I've added another task from the message on 1cfa8432f33def91cd20c18e369039b4c8c039ef.

deadtrickster:
marked prometheus_metric:value() task as complete: https://github.com/deadtrickster/prometheus.erl/commit/416fd3d131e755809839fa1f163758856567b067

[...]

PR #16, opened 2016-08-14
Replace #'Metric'{} with prometheus_model:'Metric'()

yurrriq:
Complete a task on #12.

### DIFF
--- a/src/prometheus_collector.erl
+++ b/src/prometheus_collector.erl
@@ -11,8 +11,6 @@
 
 -compile({no_auto_import, [register/2]}).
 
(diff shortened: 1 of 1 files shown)


deadtrickster:
Why elvis doesn't complain about No Spec With Records rule violation - https://github.com/inaka/elvis_core/wiki/Rules#no-spec-with-records?

yurrriq:
> Why elvis doesn't complain about No Spec With Records rule violation...?

That's a great question...

deadtrickster:
Still need to figure out behavior differences

PR #19, opened 2016-08-15
Update gpb config

yurrriq:
- Drop runtime dependency by adding `defs_as_proplists` to config
- Add `clean` post-hook
- Normalize `rebar.config`, while we're at it


rebar.config
```diff
@@ -1,14 +1,12 @@
- % -*- mode: Erlang;-*-
-{erl_opts, [debug_info, {i,"./_build/default/plugins/gpb/include/"}]}.
+%% -*- mode: erlang -*-
+{erl_opts, [debug_info]}.
 {plugins, [{rebar3_gpb_plugin, "1.7.2"}]}.
-{provider_hooks, [
-                  {pre, [{compile, {protobuf, compile}}]}
-                 ]}.
+{provider_hooks, [{pre,  [{compile, {protobuf, compile}}]},
+                  {post, [{clean,   {protobuf, clean}}]}]}.
```
deadtrickster: this will remove generated .erl & .hrl for model.proto whenever rebar3 clean executed, right?

yurrriq: Exactly.

deadtrickster:
I was writing description for the 2.0.0 tag when I saw this =).
Eliminating compile-time dependency was actually one of the main goals. But I forgot to add issue for it :-(.
It's important because erlang.mk doesn't work with rebar3 compile plugins well and since RabbitMQ is erlang.mk based building plugins is extremely painful. 
Thank you!

Issue #4, opened 2016-07-24
Write Documentation


deadtrickster:
Can hex.pm host edoc generated docs?

Checklist for modules:
- [x] buckets;
- [x] collector;
- [x] format;
- [x] registry;
- [x] vm_memory_collector;
- [x] vm_statistics_collector;
- [x] vm_system_info_collector;
- [x] http;
- [x] protobuf_format;
- [x] text_format;
- [x] counter;
- [x] gauge;
- [x] histogram;
- [x] summary;
- [x] model_helpers.

Checklist for pages:
 - [x] Overview;
 - [x] Time intervals.

yurrriq:
https://hex.pm/docs/tasks#hex_docs

yurrriq:
```erlang
{edoc_opts, [{dir, "docs"}]}.
```
+
```fish
rebar3 edoc
```

will tell [EDoc](http://erlang.org/doc/apps/edoc/chapter.html) to generate its output in `docs/`, which it sounds like https://hexdocs.pm will automatically serve.

yurrriq:
Otherwise, I quite like the GFM output of [edown](https://github.com/uwiger/edown):
```erlang
{profiles, [
  {docs, [{deps, [{edown, "0.8.1}]}]}
]}.

{edoc_opts, [
  {doclet,      edown_doclet},
  {subpackages, true}
]}.
```

PR #20, opened 2016-08-15
Add initial edoc config and docs/

yurrriq:
Generate Github Flavoured Markdown in `doc/`.

**Note**: This depends on #19, but doesn't really need to.
**Related**: #4


bin/set-edoc-hidden
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env sh
+
+[ -f "$1" ] || exit 1
+
+FILE="$1"
+HIDDEN="%% @hidden"
+
+if [ "$(head -n 1 $FILE)" != "$HIDDEN" ]; then
+    echo "0a\n$HIDDEN\n.\nw" | ed "$FILE" >/dev/null
```
yurrriq: :laughing:

yurrriq:
Rebased on master.

bin/set-edoc-hidden
deadtrickster: Nice to see you adopted bin folder!
What the purpose of this script? it adds `@hidden` tag to `FILE`?

yurrriq:
As noted in #4, we can use

```erlang
{edoc_opts, [{dir, "docs"}]}.
```

to generate (probably) [hexdocs](https://hexdocs.pm)-friendly HTML docs in `docs/` instead. Maybe we could do this in separate profiles.

bin/set-edoc-hidden
yurrriq: If the first line is not `%% @hidden` then it prepends it.

deadtrickster:
I talked to Eric (probably it was him) and he said everything with docs/index.html is fine.

src/model/prometheus_model.erl
```diff
@@ -77,40 +78,40 @@ e_msg_Bucket(Msg, TrUserData) ->
 
 
 e_msg_Bucket(#'Bucket'{cumulative_count = F1,
-           upper_bound = F2},
-       Bin, TrUserData) ->
+		       upper_bound = F2},
```
deadtrickster: Tabs :-(

yurrriq: 3dae9f2

yurrriq:
[HTML preview](https://htmlpreview.github.io/?https://github.com/yurrriq/prometheus.erl/blob/feature/edoc-config/docs/index.html)

deadtrickster:
Why tabs sneaked anyway? Indenting prometheus_model manually doesn't look like a good idea because it auto-generated.

yurrriq:
I think I manually `untabify`'d it before too. I have save hooks that re-indent, and/or automatic key chord impulses..

yurrriq:
```fish
gsed -i 's/\t/  /g' src/model/prometheus_model.erl
```

The default macOS `sed` doesn't have `-i` IIRC. It must be GNU.

deadtrickster:
Yea I have indent-whole-buffer (bound to F12) too and I like to smash it. 
And you are right, gpb generates it with tabs. Can you please reset it 'back' to Tabs then so we wouldn't have this huge diff in the future anymore.

deadtrickster:
I update gpb/rebar3_gpb_plugin one more time. As it turned out elvis just skipped src/model :confused:.
Now elvis checks just prometheus_model_helpers. I as long as we add only `@hidden` I'm fine with modifications.

deadtrickster:
Lets delete this commit https://github.com/deadtrickster/prometheus.erl/pull/20/commits/3c664ce3b8502ce5863c30ffc15f450728bd83ca and I'll merge PR.

deadtrickster:
Thanks!

PR #34, opened 2016-09-13
Update prometheus_collector.erl

massemanet:
application:get_env/2 returns {ok,Val}


deadtrickster:
Thanks! this was fixed in version-3 branch. Give it a try it's already stable enough. If that's not an option I'll merge this. 
https://github.com/deadtrickster/prometheus.erl/releases/tag/3.0.0-alpha9

massemanet:
alpha9 is good enough for me. I'll close the pr.

PR #38, opened 2016-09-19
add mnesia collector

massemanet:
I thought this mnesia collector might be of general interest.

### DIFF
--- a/src/collectors/mnesia/prometheus_mnesia_collector.erl
+++ b/src/collectors/mnesia/prometheus_mnesia_collector.erl
@@ -1,0 +1,90 @@
+-module(prometheus_mnesia_collector).
+
+-export([deregister_cleanup/1,
+         collect_mf/2,
+         collect_metrics/2]).
+
+-import(prometheus_model_helpers, [create_mf/5,
+                                   counter_metric/1,
+                                   gauge_metric/1]).
+
+-behaviour(prometheus_collector).
+
+%%====================================================================
+%% Collector API
+%%====================================================================
+deregister_cleanup(_) -> ok.
+
+-spec collect_mf(_Registry, Callback) -> ok when
+    _Registry :: prometheus_registry:registry(),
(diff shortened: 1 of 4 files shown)


deadtrickster:
Awesome, thanks!

Could you please make exported metrics list configurable? Like here: https://github.com/deadtrickster/prometheus.erl/blob/master/src/collectors/vm/prometheus_vm_system_info_collector.erl#L325.

src/collectors/mnesia/prometheus_mnesia_collector.erl
```diff
@@ -0,0 +1,90 @@
...
+        [erlang_mnesia_failed_transactions,
+         erlang_mnesia_committed_transactions,
+         erlang_mnesia_logged_transactions,
+         erlang_mnesia_restarted_transactions]).
+-define(GAUGES,
+        [erlang_mnesia_held_locks,
+         erlang_mnesia_lock_queue,
+         erlang_mnesia_transaction_participants,
+         erlang_mnesia_transaction_coordinators,
+         erlang_mnesia_current_transactions]).
```
deadtrickster: can't find where value comes from

massemanet: participants and coordinators come from the  transaction manager (mnesia_tm:get_info).
lock_queue and held_locks come from looking at the locks ets tables (to avoid reading out all the locks from the tables and looking at the length).
erlang_mnesia_current_transactions doesn't exist; it's a bug...

deadtrickster: yes, yes I meant current_transactions. this is why I test presence of the all metrics in my collectors :-)

deadtrickster:
Maybe collect information from mnesia events too? http://erlang.org/doc/apps/mnesia/Mnesia_chap5.html#event_handling

deadtrickster:
Thank you!

PR #40, opened 2016-09-30
fix hash bang doesn't work on windows

deadtrickster:
See https://github.com/deadtrickster/prometheus.erl/issues/39.

@yurrriq could you please confirm this still works on OS X?


yurrriq:
Still works for me on macOS 10.12.

PR #41, opened 2016-09-30
Regenerate docs

yurrriq:
Delete generated Markdown docs that are marked `@hidden` and regenerate
`doc/prometheus/mnesia_collector.md` (to use Erlang-style atoms instead
of Elixir-style `:`-prefixed ones) by running `rebar3 edoc`.


deadtrickster:
Thanks!

PR #45, opened 2016-10-25
collected counter can be NaN

massemanet:
A counter value can actually be 'undefined', but the spec claims it can not. This confuses dialyzer.


deadtrickster:
Thank you!
I think we already have `value()` type: https://github.com/campanja-forks/prometheus.erl/blob/8e46c04c6775284081176343b1be5cec9457557e/src/model/prometheus_model_helpers.erl#L44. 
As of now it is used by Gauge, Untyped and other metrics. Maybe we could use it here too?

massemanet:
good suggestion

deadtrickster:
Thanks!

PR #49, opened 2016-11-05
Windows set-edoc-hidden script

deadtrickster:
related: https://github.com/deadtrickster/prometheus.erl/pull/46, https://github.com/deadtrickster/prometheus.erl/issues/39


yurrriq:
Nice!

deadtrickster:
Not needed really anymore, since file won't be updated/generated in default rebar3 profile

PR #51, opened 2016-12-08
allow prometheus.erl to be used as a dep in mix projects

talentdeficit:
this adds an application start (mod) key and adds mix to the package metadata to allow prometheus to be used as a dependency from mix


deadtrickster:
Hi, did you have a chance to try prometheus.ex?

deadtrickster:
Closing this, Please let me know if prometheus.ex didn't work

PR #63, opened 2017-05-03
Portable float NIF

deadtrickster:
Replace #62 with more portable nif (courtesy of @benoitc)
(Extracted from https://gitlab.com/barrel-db/instrument, `Gauge` renamed to `Value`, `instrument` namespace to `prometheus` and target renamed from `instrument_nif` to `prometheus_nif`)

**TODO**

- [ ] Limit number of iterations to prevent long-running or endless? CAS loops
- [ ] Implement prometheus_value which tries to load this nif but fallbacks to gen_server (Windows)
- [ ] Make compilation conditional via envvar (we don't need nif troubles with rabbitmq plugin-like use cases, where we don't instrument directly)
- [ ] Adjust makefile targets + finally add rebar3_archive_plugin

Thanks @benoitc  ;-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 sudo: false
+dist: trusty
 language: erlang
 install: 'true'
 before_script:
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ rebar3.crashdump
 _*
 .eunit
 *.o
+*.so
 *.beam
 *.plt
 *.swp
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -1,4 +1,5 @@
 sudo: false
+dist: trusty
```
deadtrickster: default travis vms are very old (

NelsonVides:
Closing as won't be needed, OTP floats has gotten so much faster anyways :)

Issue #64, opened 2017-05-11
prometheus_http module conflict with the one in prometheus_httpd


benoitc:
Both `prometheus` and `prometheus_httpd` contain a `prometheus_http` module which conflict at build time.

PR #65, opened 2017-05-11
rename prometheus_http to prometheus_http_helper

benoitc:
this change make sure it won't conflict with prometheus_httpd


benoitc:
related to #64

PR #70, opened 2017-09-27
Make it work with rebar2. close #69

### LINKED ISSUE #69: rebar2 support
Hi, I'm working on a project that's currently using rebar2, and I had some minor trouble adding prometheus.erl in without modification. (I would migrate to rebar3, but it is non-trivial, I want to add metrics/monitoring *first*.)

with just a fresh checkout of prometheus.erl, the following happens:
```
# rebar compile
==> prometheus.erl (compile)
ERROR: Protobuffs library not present in code path!
ERROR: compile failed while processing /<path>/prometheus.erl: rebar_abort
```

This stems from the fact that rebar is trying to generate erlang source from `src/model/prometheus_model.proto`, but the library is not installed.  Since `src/model/prometheus_model.erl` already exists and whatever generate step is not necessary for things to work, my solution was `rm src/model/prometheus_model.proto` and then everything works fine, wrapped in a Makefile.

I tried to ignore or delete the file from rebar, but couldn't find a way to do that when it's used as a dependency.

It would be nice if I didn't have to wrap it with a Makefile.

deadtrickster:
cc @Stelminator

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,9 @@ language: erlang
 install: 'true'
 before_script:
 - wget https://s3.amazonaws.com/rebar3/rebar3
+- wget https://github.com/rebar/rebar/wiki/rebar
 - chmod +x rebar3
+- chmod +x rebar
 env: PATH=$PATH:.
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


Stelminator:
@deadtrickster confirmed, your [`rebar2` branch](https://github.com/deadtrickster/prometheus.erl/tree/rebar2) fixes https://github.com/deadtrickster/prometheus.erl/issues/69 for me.

PR #75, opened 2018-05-01
Collect allocators

essen:
For all "util" allocators we export the blocks, blocks_size, carriers, carriers_size for the sbcs, mbcs and mbcs_pool of each allocators. This can be used to help tweak the various memory allocator options of the VM.

We will submit a dashboard based on this to https://github.com/deadtrickster/beam-dashboards later today.

cc @gerhard

### DIFF
--- a/src/collectors/vm/prometheus_vm_system_info_collector.erl
+++ b/src/collectors/vm/prometheus_vm_system_info_collector.erl
@@ -317,8 +317,8 @@ allocators() ->
     Allocators = erlang:system_info(alloc_util_allocators),
     %% versions is deleted in order to allow the use of the orddict api,
     %% and never really having come across a case where it was useful to know.
-    [{{A,N},lists:sort(proplists:delete(versions,Props))} ||
+    [{{A, N}, lists:sort(proplists:delete(versions, Props))} ||
         A <- Allocators,
-        Allocs <- [erlang:system_info({allocator,A})],
+        Allocs <- [erlang:system_info({allocator, A})],
         Allocs =/= false,
-        {_,N,Props} <- Allocs].
+        {_, N, Props} <- Allocs].
--- a/src/collectors/vm/prometheus_vm_system_info_collector.erl
+++ b/src/collectors/vm/prometheus_vm_system_info_collector.erl
@@ -304,7 +304,7 @@ collect_allocator_metrics() ->
             [
                 allocator_metric(Alloc, Instance, Kind, Key, KindInfo)
             || Key <- [blocks_size, carriers_size]]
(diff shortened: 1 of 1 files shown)


deadtrickster:
Cool! could you please look at this: https://travis-ci.org/deadtrickster/prometheus.erl/jobs/373386336#L529

essen:
Done.

deadtrickster:
tagged/packaged as 4.2.0

gerhard:
How safe is to use prometheus.erl v4.2.0 in combination with rabbitmq_prometheus_exporter v3.7.2.1 & v3.6.14 ?

deadtrickster:
if you run on OTP > 20 it's safe, if you don't let me know I'll merge this to 3.x

gerhard:
Ideally, it should work with OTP 19.3 and above. Since OTP 21 is close, prometheus.erl could (should?) be OTP n-2 compatible.

Am I correct in assuming that rabbitmq_prometheus_exporter v3.6.14 works with OTP R16B03 and above, and rabbitmq_prometheus_exporter v3.7.2.1 works with OTP 19.3 and above? This is my reference: [RabbitMQ Erlang Version Requirements](http://www.rabbitmq.com/which-erlang.html)

deadtrickster:
Yes you are right, because v3.7 uses prometheus < 4. I tagged 3.5 with these allocator commits added. also 3.5 is on hex now. It could take some time to update ez in the exporter release. rebar3 archive will work here too.

gerhard:
I got 3.5 working, but I'm getting the followign error for prometheus_rabbitmq_exporter v3.7.2.1 on Erlang/OTP v19.3.6.8:

```
{plugin_built_with_incompatible_erlang,"prometheus_rabbitmq_exporter"}
```
I am thinking to `rebar3 archive` all prometheus plugins explicitly. Does this look correct?

| | RabbitMQ 3.6.x | RabbitMQ 3.7.x |
| - | - | - |
| Erlang/OTP | v19.3.6.8 | v20.3.5 |
| accept | v0.3.3 | v0.3.3 |
| prometheus.erl | v3.5.0 | v3.5.0 |
| prometheus_rabbitmq_exporter | v3.7.2.1 | v3.6.14.1 |
| prometheus_httpd | v2.1.8 | v2.1.8 |
| prometheus_cowboy | N/A | v0.1.4 |

Also, how would you recommend creating `accept-0.3.3.ez` against the correct Erlang/OTP version?

deadtrickster:
prometheus_rabbitmq_exporter looks swapped. `accept.ez` can be created like the rest. In the worst case rebar3 archive plugin can be added to rebar.config. Loic has my buildable archive of prometheus_rabbitmq_exporter btw.

Issue #78, opened 2018-05-18
Allocators instance label clashes with native Prometheus label


deadtrickster:
I somehow missed it but erlang_vm_allocators instance label conflicts with autoassigned. it could be renamed to instance_no or alloctor_no. 

By default prometheus renames this label to exported_instance and replace with scrape  target. When honor_lables is true it doesn't (this is what we use).

cc @gerhard

essen:
I can send a patch renaming from `instance` to `instance_no`. Better do it now before we create too many graphs out of it.

deadtrickster:
Thank you Loïc

PR #79, opened 2018-05-29
Rename instance to instance_no to avoid conflict

essen:
https://github.com/deadtrickster/prometheus.erl/issues/78


essen:
@gerhard tells me we are using `v3` (3.5.0), can you cherry-pick it there or do you want a separate PR?

deadtrickster:
A separate PR would be cool

PR #80, opened 2018-05-29
Rename instance to instance_no to avoid conflict

essen:
Same as https://github.com/deadtrickster/prometheus.erl/pull/79 but for the v3 branch.

https://github.com/deadtrickster/prometheus.erl/issues/78


deadtrickster:
Thanks  :-)

PR #90, opened 2019-04-16
Add erlang_vm_wordsize to system_info collector

essen:
Warning: I have not tested it yet.

This value is useful in order to convert values to/from bytes and compare them. There is currently one value in words in the prometheus.erl collectors and it is also available in bytes so this wasn't really necessary. But I am working on adding a collector that has multiple word values and would prefer to avoid providing one value per unit and converting on the collector's side. I think it's best left out for Grafana or other to do when necessary.

### DIFF
--- a/src/collectors/vm/prometheus_vm_system_info_collector.erl
+++ b/src/collectors/vm/prometheus_vm_system_info_collector.erl
@@ -95,6 +95,11 @@
 %%     1 if time correction is enabled, otherwise 0.
 %%   </li>
 %%   <li>
+%%     `erlang_vm_wordsize_bytes'<br/>
+%%     Type: gauge.<br/>
+%%     The size of Erlang term words in bytes.
+%%   </li>
+%%   <li>
 %%     `erlang_vm_atom_count'<br/>
 %%     Type: gauge.<br/>
 %%     The number of atom currently existing at the local node.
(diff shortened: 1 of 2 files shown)


gerhard:
Nice one! Happy to run a preliminary test shortly 👍

gerhard:
@essen which metric is in words?

essen:
```
# TYPE erlang_vm_statistics_garbage_collection_bytes_reclaimed counter
# HELP erlang_vm_statistics_garbage_collection_bytes_reclaimed Garbage collection: bytes reclaimed.
erlang_vm_statistics_garbage_collection_bytes_reclaimed 236563328
# TYPE erlang_vm_statistics_garbage_collection_words_reclaimed counter
# HELP erlang_vm_statistics_garbage_collection_words_reclaimed Garbage collection: words reclaimed.
erlang_vm_statistics_garbage_collection_words_reclaimed 29570416
```

deadtrickster:
please rebase, should be green? also, for the test I think I just match the string

gerhard:
Looking good:

image

essen:
Added tests and fixed a small thing in the docs. Also rebased.

gerhard:
The name should be `erlang_vm_wordsize_bytes`, so it's missing just `_bytes` and the end. Looking good otherwise!

essen:
Done.

gerhard:
This is good to merge.

What do you think @deadtrickster, have @essen & myself earned contributor privileges by now?

PR #94, opened 2019-06-11
Add node_queue_size metric to dist collector

essen:
This depends on https://github.com/erlang/otp/pull/2270 to be really useful. Wait before merging as we're still experimenting.

It's compatible with non-patched OTP versions but the value is simply not useful in that case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,6 @@ cache:
   directories:
   - $HOME/.cache/rebar3/
 otp_release:
-- 20.1
-- 20.0
+- 21.3
+- 22.3
 script: "./bin/checks.sh && ./rebar3 as test coveralls send"
--- a/doc/prometheus_vm_dist_collector.md
+++ b/doc/prometheus_vm_dist_collector.md
@@ -128,6 +128,10 @@ The current state of the distribution link.<br />
 The state is represented as a numerical value where `pending=1`,
 `up_pending=2` and `up=3`.
 
+* `erlang_vm_dist_node_queue_size_bytes`<br/>
+Type: gauge.<br/>
+The number of bytes in the output distribution queue.<br/>
+This queue sits between the Erlang code and the port driver.
 
 
 ### <a name="Configuration">Configuration</a> ###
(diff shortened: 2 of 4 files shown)


essen:
I will need to rename the metric and fix its description because it is in bytes.

essen:
@gerhard Since the OTP PR got merged we can finish and merge this next week, just need to remember to rename the metric and update its description.

gerhard:
@essen since 22.1 & 22.1.1 have been released, should we wrap this up?

essen:
Sure but don't wait for me it should be trivial to finish up.

gerhard:
OK, will re-submit as a new PR.

essen:
@gerhard What name did we decide to use?

gerhard:
I would go with `node_link_queue_messages_bytes` since it's the size in bytes of all messages queued in the Erlang Distribution link to another peer node.

essen:
Done.

essen:
I still need to update the description.

essen:
This will not pass the xref checks in Travis because the function we use was only added in OTP-21.0 apparently. How do you want me to solve this?

gerhard:
We should definitely bump the Erlang versions in `.travis.yml`.

I advocate for testing this just on OTP 21 & 22, basically n-1. FWIW, [RabbitMQ soft-deprecated OTP 20.x support in August 2019](https://groups.google.com/forum/#!searchin/rabbitmq-users/ANN|sort:date/rabbitmq-users/9tc_OE1eMPk/ly1NEISwBwAJ) & we no longer support this OTP major since 3.8.0 (shipped October 2019).

If we decide to keep OTP 20 support, I would add a conditional to disable xref checking for this version.

What do you think @deadtrickster ?

essen:
The README says support is for R16B03+. I don't know how accurate that is.

gerhard:
@deadtrickster what OTP support policy would you think is appropriate for prometheus.erl? I would recommend `n-1`, meaning OTP 21 & 22 currently, and OTP 22 & 23 once the new major ships in June.

The reason why I ask is because this change will not work on OTP 20, as there is no `erlang:dist_get_stat/1`.

deadtrickster:
I'm happy with n-1

gerhard:
@deadtrickster my 🍿is out, in anticipation of the next 4.x release 😉

PR #96, opened 2019-07-10
Allow multiple arities metrics

jfacorro:
This PR enables registering and reporting a different number of labels for the same metric (which is supported by metrics services like Datadog). 

The resulting metrics in the prometheus generated page would be:
```
# TYPE foo gauge
# HELP foo 
foo{bar="a"} 1
# TYPE rimu_serv
# TYPE foo gauge
# HELP foo 
foo{bar="a",baz="b"} 2
```


jfacorro:
Here is a snippet of Erlang code that does the declaring, incrementing and value getting of metrics with a different amount of labels:
```
prometheus_counter:declare([{name, foo}, {help, "foo"}]).
prometheus_counter:declare([{name, foo}, {help, "foo"}, {labels, [bar]}]).
prometheus_counter:declare([{name, foo}, {help, "foo"}, {labels, [bar, baz]}]).

prometheus_counter:inc(foo).
prometheus_counter:inc(foo, [a]).
prometheus_counter:inc(foo, [a]).
prometheus_counter:inc(foo, [a, b]).
prometheus_counter:inc(foo, [a, b]).
prometheus_counter:inc(foo, [a, b]).

prometheus_counter:value(foo). %%= 1
prometheus_counter:value(foo, [a]). %%= 2
prometheus_counter:value(foo, [a, b]). %%= 3
```
This is how the prometheus metrics page show them:
```
# TYPE foo counter
# HELP foo foo
foo 1
# TYPE foo counter
# HELP foo foo
foo{bar="a"} 2
# TYPE foo counter
# HELP foo foo
foo{bar="a",baz="b"} 3
```

I hope I understood correctly what you where asking for.

deadtrickster:
Yes that's exactly what I wanted to see. 

But there is also: https://prometheus.io/docs/instrumenting/writing_clientlibs/#labels which says 

```
Client libraries MUST NOT under any circumstances allow users to have different label names for the same metric for Gauge/Counter/Summary/Histogram or any other Collector offered by the library
```

:man_shrugging:

jfacorro:
I have been thinking if there can be some kind of compromise between the prometheus specs and this requirement we have to use different amount of labels for the same metric, but it has not been very fruitful.

I am curious if you have never heard of people wanting the feature added in the PR. We report metrics in different parts of our code and it might be that extra information is useful (and only available) in certain contexts (e.g. include the `reason` for an `error`, or the `result` value for a call).

Since this PR goes against to what the prometheus specs state, I see no other alternative than to close it without merging. 😞

Thank you for checking it out though.

deadtrickster:
>. We report metrics in different parts of our code and it might be that extra information is useful (and only available) in certain contexts (e.g. include the `reason` for an `error`, or the `result` value for a call).

I would have created two metrics in this case. Like requests_count{result=error|value} and errros_count{error=error}.

For cases when label value is unknown one can always use undefined. Or we can add default label value. Not sure about labels ordering but say accepting plist/dict/hashtable would work here.

PR #102, opened 2020-01-30
Support Erlang 23 format for blocks entry in allocators

dcorbacho:
`blocks_size` has been removed, instead the `count` and `size` info is provided in `blocks`

```
{blocks,[{ll_alloc,[{count,1,1,1},{size,131144,131144,131144}]}]}
```

### DIFF
--- a/src/collectors/vm/prometheus_vm_system_info_collector.erl
+++ b/src/collectors/vm/prometheus_vm_system_info_collector.erl
@@ -315,7 +315,11 @@ collect_allocator_metrics() ->
         [
             [
                 allocator_metric(Alloc, Instance, Kind, Key, KindInfo)
-            || Key <- [blocks, blocks_size, carriers, carriers_size]]
+            || Key <- [carriers, carriers_size]] ++
+            [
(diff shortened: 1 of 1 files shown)


deadtrickster:
as far as I can tell it's meant to be backwards compatible, right? Also build fails

essen:
Yes that's what I meant, it's backward compatible. And yes, build fails. :-)

dcorbacho:
Thank you both, it's fixed now. I hadn't compiled it independently from rabbit..

deadtrickster:
still red :-(

essen:
Coverage necessarily decreases because the new code will only run in OTP 23.

deadtrickster:
Thanks!

gerhard:
I am finally catching up on this. I've noticed that v4.5.0, the latest release currently available, [does not include this change](https://github.com/deadtrickster/prometheus.erl/compare/v4.5.0...f9c7a221b280577c9a1a5fb6c036d72d2dfb9390).

If this is backwards-compatible with OTP 22 & 21, are we waiting for anything else before we cut a new release? There is no rush from my perspective, just wondering.

PR #105, opened 2020-02-24
Add DIAMETER Exporter to README.md

RoadRunnr:


deadtrickster:
Thanks!

RoadRunnr:
Actually, that collector is so simple, do you rather want it to be included with the main project?
I could make a PR for that (although without unit tests).

deadtrickster:
mmm, no unit tests... would be awesome if you throw something...

PR #107, opened 2020-05-07
add metric for mnesia memory usage

ipinak:
This PR adds a counter metric to show the accumulated memory usage of mnesia tables.

### DIFF
--- a/doc/prometheus_mnesia_collector.md
+++ b/doc/prometheus_mnesia_collector.md
@@ -50,6 +50,9 @@ Number of transactions logged.
 Type: counter.<br />
 Total number of transaction restarts.
 
+* `erlang_mnesia_memory_usage`<br/>
+Type: counter.<br/>
+Total number of bytes allocated by all mnesia tables.
 
 
 ### <a name="Configuration">Configuration</a> ###
@@ -66,5 +69,6 @@ Available options:
 - `transaction_commits` for `erlang_mnesia_committed_transactions`;
 - `transaction_log_writes` for `erlang_mnesia_logged_transactions`;
 - `transaction_restarts` for `erlang_mnesia_restarted_transactions`.
+- `memory_usage` for `erlang_mnesia_memory_usage`.
 
 By default all metrics are enabled.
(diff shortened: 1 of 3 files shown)


essen:
The metric name should include the unit, for example `erlang_mnesia_memory_usage_words` or `erlang_mnesia_memory_usage_bytes`.

There is `erlang_vm_wordsize_bytes` https://github.com/deadtrickster/prometheus.erl/commit/fd2a750e5dc8051459f773bf643a34847d0003e4 so perhaps you shouldn't do the words to bytes conversion when exporting and just do it in diagrams or other.

/cc @gerhard

gerhard:
Agreed, `erlang_mnesia_memory_usage_bytes` makes most sense.

See the [Prometheus documentation on metric & label naming](https://prometheus.io/docs/practices/naming/) for context.

ipinak:
> The metric name should include the unit, for example `erlang_mnesia_memory_usage_words` or `erlang_mnesia_memory_usage_bytes`.
> 
> There is `erlang_vm_wordsize_bytes` [fd2a750](https://github.com/deadtrickster/prometheus.erl/commit/fd2a750e5dc8051459f773bf643a34847d0003e4) so perhaps you shouldn't do the words to bytes conversion when exporting and just do it in diagrams or other.
> 
> /cc @gerhard

If I don't do the conversion to bytes, should the metric be named `erlang_mnesia_memory_usage_words` instead?

gerhard:
Yes, if the value was no converted to bytes, then `_words` would have been the correct suffix.

The `word` unit however would make it harder to visualise it in e.g. Grafana, because there is no such data metric unit. `_bytes` is spot-on!

doc/prometheus_mnesia_collector.md
```diff
@@ -50,6 +50,9 @@ Number of transactions logged.
 Type: counter.<br />
 Total number of transaction restarts.
 
+* `erlang_mnesia_memory_usage_bytes`<br/>
+Type: counter.<br/>
```
gerhard, line 54: I believe this is a [gauge](https://prometheus.io/docs/concepts/metric_types/#gauge), not a [counter](https://prometheus.io/docs/concepts/metric_types/#counter). If you agree, can you please update it?

Happy to QA & merge afterwards 😉

ipinak: You are absolutely right, a gauge is more suitable.

PR #113, opened 2020-09-11
Tablewise Mnesia Metrics ... Optimized collect_metric for summaries and better metric registrations

SoniCoder:
This PR introduces following enhancements/fixes:

1. Tablewise Mnesia Metrics: We wished to put alerts on per table row size / mem size and thought it will be a good contribution for everyone.

2. Optimized collect_metric for summaries which was taking long time for many series of summaries (~9s), after optimization it completes under 500ms using O(N) instead of earlier O(N^2) implementation.

3. Better Metric Registration: If prometheus app is restarted then all metrics all lost and caller apps start crashing because those metrics are no longer defined. Using prometheus_sup:register_metrics, they are persisted in environment vars so upon restart they are reinitialized automatically.

### DIFF
--- a/src/metrics/prometheus_summary.erl
+++ b/src/metrics/prometheus_summary.erl
@@ -294,12 +294,21 @@ values(Registry, Name) ->
       DU = prometheus_metric:mf_duration_unit(MF),
       Labels = prometheus_metric:mf_labels(MF),
       MFValues = load_all_values(Registry, Name),
-      [begin
-         {Count, Sum} = reduce_label_values(LabelValues, MFValues),
-         {lists:zip(Labels, LabelValues), Count,
-          prometheus_time:maybe_convert_to_du(DU, Sum)}
-       end ||
-        LabelValues <- collect_unique_labels(MFValues)]
+      ReducedMap = lists:foldl(
+        fun([L, C, IS, FS], ResAcc) ->
+          {PrevCount, PrevSum} = maps:get(L, ResAcc, {0,0}),
+          ResAcc#{L => {PrevCount + C, PrevSum + IS + FS}}
+        end,
+      #{},
+      MFValues),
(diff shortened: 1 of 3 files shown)


src/metrics/prometheus_summary.erl
```diff
@@ -322,13 +331,22 @@ collect_mf(Registry, Callback) ->
...
-   end ||
-    LabelValues <- collect_unique_labels(MFValues)].
+  ReducedMap = lists:foldl(
+        fun([L, C, IS, FS], ResAcc) ->
+          {PrevCount, PrevSum} = maps:get(L, ResAcc, {0,0}),
+          ResAcc#{L => {PrevCount + C, PrevSum + IS + FS}}
+        end,
+      #{},
+      MFValues),
+  ReducedMapList = lists:sort(maps:to_list(ReducedMap)),
```
deadtrickster, line 341: why this sort needed?

deadtrickster (changes requested):
There are minor formatting issues

deadtrickster:
Hi, cool MR, thank you. What lead to prometheus app crashing?

src/metrics/prometheus_summary.erl
SoniCoder: I included this sort to output data in the same order as before I.e. Sorted so they can be easily found when accessing the API but theoretically we can also remove this sort if it is acceptable

SoniCoder:
> Hi, cool MR, thank you. What lead to prometheus app crashing?

The consequence of the crash I.e. Metric tables missing was reported by our developers so I don't exactly know what led to the app crashing

deadtrickster (approved):

essen (approved):
Gave just a very quick look.

deadtrickster:
Thank you

PR #115, opened 2021-01-08
implement running and total mnesia node count metrics

luos:
Hi,

It would be useful for us to expose running and total mnesia node counts.
Let me know what you think.

Thanks


deadtrickster:
Hi, I'm somewhat concerned with arity here. can peer label explode?

luos:
Hi,

I guess, that is a a good question. In our typical installation it's about a maximum of 7 nodes connected in an mnesia cluster. It should be same or less than the `node_state` metric coming from the `dist` collector (if all nodes online). 

I am not sure if others use many more mnesia nodes?

If you have many nodes, let's say the mentioned 7, that would create 7*6 metrics in total, so I guess you can get somewhat flooded by these.

luos:
Hi,

Do you think this can be merged, should I change something? Do you still have concerns?

Thanks

Issue #18, opened 2016-08-15
Full summary implementation


deadtrickster:
Now only basic variant of summary metric is implemented - with sum and count of observations.
Need to implements
- [ ] Quantiles;
- [ ] Configurable sliding window.

hairyhum:
So I found this: https://github.com/odo/quantile_estimator
It's rather old but seems to build, at least on OTP-21. 
I'll take a look if it's possible to integrate to the summary metric.

NelsonVides:
Quantiles redone here: https://github.com/prometheus-erl/prometheus.erl/pull/177 Note that the biased quantiles from Cormode were not mergeable and the previous implementation was crashing. See PR for details.

We don't have sliding windows though (and I doubt we want them), so I'm not sure if closing this yet, will leave it open for a while for reference.

PR #116, opened 2021-02-05
Add quantiles to summary using quantiles_estimator package.

hairyhum:
Trying to address #18 
Using https://github.com/odo/quantile_estimator to calculate quantiles.

The merge function is not ideal, but works well enough on simple samples.

In terms of backwards compatibility I can create a separate module for summaries with quantiles or add a config to be able to enable/disable them.


deadtrickster:
Hi, thank you for looking into this. I will be more comfortable if summary with quantiles will be put in the new module.

deadtrickster:
@essen @gerhard what you think?

essen (commented):
I'll let @gerhard deal with that one, it's too "metrics-technical" for me.

gerhard:
Nice to bump into you @hairyhum 🙂

This does look useful & I like the clean separation - good call @deadtrickster.

I am somewhat concerned about the lack of change in https://github.com/odo/quantile_estimator. Not a problem in itself, but do we know if it works on OTP 23? How about OTP 24? While 24 is not GA yet - [RC1 shipped a few weeks ago](https://github.com/erlang/otp/releases/tag/OTP-24.0-rc1) -, I am wondering how much work there will be to get this new library working in this very important upcoming version.

Do we know why the Travis CI builds are failing?

hairyhum:
I'm running my tests on OTP23. I can run with 24RC as well.

gerhard:
OK, I am assuming that all tests are passing on 23.

Yes please for 24.

hairyhum:
I ran a quick test on 24.0-rc1 with `rebar3 eunit` and there is no issues other than compile warnings about bifs.

deadtrickster:
Merged, thanks! I will let it live in master for some time before cutting release, want to play with 24 myself.

PR #124, opened 2021-04-29
improve efficiency of label translation in text format

RoadRunnr:
Optimize label to string translation to avoid redundant work.

Experiments show a up to 50% improvement on large datasets.


deadtrickster:
Thank you for the patches!

PR #108, opened 2020-05-12
Added functions `clear/1,2,3` to all metrics

balena:
This PR adds functions `clear/1,2,3` to all metrics allowing users to clear (erase) existing metrics. This is useful case the app needs to stop (interrupt) certain metrics along its lifecycle.

PR #125, opened 2021-05-19
Added functions match_remove/1,3,4 to all metrics

RoadRunnr:
match_remove can be used to selectively remove values for multiple
labels. The match head and match condition use the same rules as the
match specification parts of the same name.

This similar to #108, but can use arbitrary match conditions to select which labels to remove.


RoadRunnr:
@deadtrickster can you take a look?

mgumz:
@deadtrickster ping

PR #126, opened 2021-06-08
RabbitMQ license has changed to MPL2

michaelklishin:
so license header in `prometheus_misc` has to be updated.

This shows up as a transient MPLv1.1 dependency on VMware open source
compliance reviews.

### DIFF
--- a/src/prometheus_misc.erl
+++ b/src/prometheus_misc.erl
@@ -1,20 +1,12 @@
 %% @hidden
 %% behaviour_modules original Copyright message
 %% all other code is under MIT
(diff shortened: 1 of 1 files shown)


michaelklishin:
@deadtrickster I know this is silly but would it be possible to produce a new patch release with this change?

deadtrickster:
yep, git tag or hex package or both?

michaelklishin:
We depend on a Hex version I believe. Obviously a tag would not hurt.

deadtrickster:
4.8.1 should be available

PR #136, opened 2021-10-13
Write bigger chunks to ram_file

binarin:
`ram_file` is implemented as a port driver, and has quiet a significant overhead if every single series is emitted separately.

What happens here is that bigger chunk is collected before writing to ram_file. Chunk is collected using binary append, in such a way that is can benefit from binary append optimization in ERTS.

One nice side-effect of collecting binaries is that a lot of iolists in the hottest code path are no longer created, putting significantly
less pressure on the garbage collector.

Observed performance gains can be as high as 20 times (700k metrics emitted by RabbitMQ went from 2 minutes to 5 seconds). Such big difference can be mostly attributed to garbage collection overhead (for some reason it's non-linear in the number of metrics), but profiling shows that 20% improvement is there even without accounting for gc.

### DIFF
--- a/include/prometheus.hrl
+++ b/include/prometheus.hrl
@@ -19,4 +19,4 @@
         error_logger:warning_msg(Old " is deprecated and will soon be removed. "
                                  "Please use " New " instead.~n")).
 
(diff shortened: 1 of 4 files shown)


deadtrickster:
@binarin please look at the pipeleines - some minor code style things

PR #141, opened 2022-06-07
Update edown 0.8.4

mopp:
## Why

```console
❯ erlenv release
24.2 (set by /Users/mopp/.anyenv/envs/erlenv/release)

> erl
Erlang/OTP 24 [erts-12.2] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit]

Eshell V12.2  (abort with ^G)
1>
```

I've used Erlang/OTP 24.2 and I found `rebar3 edoc` causes the error below because I tried to fix the documents.
See also https://github.com/deadtrickster/prometheus.erl/pull/142

```console
> ./_build/rebar3 edoc
===> Verifying dependencies...
===> Analyzing applications...
===> Compiling edown
===> Compiling _build/docs/lib/edown/src/edown_doclet.erl failed
_build/docs/lib/edown/src/edown_doclet.erl:112:15: record context undefined
_build/docs/lib/edown/src/edown_doclet.erl:113:15: record context undefined
_build/docs/lib/edown/src/edown_doclet.erl:114:20: record context undefined
```

## What I did

I just update the down version and regenerate the document.

PR #142, opened 2022-06-07
Fix links to Erlang/OTP official document

mopp:
Please review and merge https://github.com/deadtrickster/prometheus.erl/pull/141 in advance.

## Overview

Some links are suffixed by edown
e.g., `http://erlang.org/doc/man/mnesia.md#system_info-1`

We should use `http://www.erlang.org/doc/man/` instead of `http://erlang.org/doc/man/` because the edown recognizes the former. Otherwise it replaces `.html` to `.md`.
https://github.com/uwiger/edown/blob/754be25f71a04099c83f3ffdff268e70beeb0021/src/edown_lib.erl#L44-L52


NelsonVides:
Closing for the same reason than #141.

PR #143, opened 2022-07-08
Prevent deadlock with application_controller shutdown

slezakattack:
When an application is shutting down, calling `application:which_applications()` causes a deadlock as it waits for a response from the already-busy `application_controller` process. This is preventing my project from shutting down gracefully due to our metrics scraper requesting metrics while the service is shutting down. The easiest way to check if mnesia is running without using the `application_controller` is to call `mnesia:system_info(is_running)`.

I was trying to write a test for this but was having difficulty simulating a "dummy application" in eunit. I'm open to suggestions on this if a test is strictly required.


deadtrickster:
Thanks!

PR #147, opened 2023-01-09
Support global and prerendered labels together

mkuratczyk:
Without this change, a collector that pre-renders labels (like RabbitMQ) will return `<<>>` in case there are no labels at all. That will create an an improper list when an empty binary is appended to a list of global labels. Rather than dealing with improper list later, let's just avoid them.


binarin:
It's better to pre-render global labels once to a single binary, outside the callback function - otherwise relatively large set of global labels can crash performance even if the app pre-renders labels itself.

Issue #81, opened 2018-06-26
Slow Supervisor Boot Time


lswith:
Context:
I've added the prometheus_ex library to my phoenix application. When I do this, my mix tasks and overall boot time for the whole application has gone up by a minute. I've started trying to determine where this occurs, and it's definitely happening the moment this application is added to the supervision tree.

Any thoughts on how I can improve runtime boot time?

deadtrickster:
This might happen because of collectors autodiscovery. To check that you can configure collectors list manually (note default alias) https://github.com/deadtrickster/prometheus.erl#configuration

lswith:
I love the quick response!

I've actually set the default_collectors to an empty list to test the boot time performance and its still very similar.

lswith:
So, setting the instrumenters to an empty list as well as the collectors solved the boot time issue. Any ideas what might be the reason for this?

deadtrickster:
Not sure, what you mean by instrumenters?

lswith:
you can configure `config :prometheus, instrumenters: []`.

lswith:
I think the issue might be in this module: https://github.com/deadtrickster/prometheus.erl/blob/master/src/prometheus_misc.erl

Getting the default instrumenters seems very slow.

lswith:
I'm just curious what the default instrumenters are? if I wanted to set this configuration to the default, what would it be?

deadtrickster:
hmm looks like this is a leftover from the past, I was trying to automate instrumenters set up like declaring metrics, but due interactive/embedded mode differences it's not possible to do that reliably. As a temporary fix please add instrumenters: [] to the prometheus config. I'll remove it entirely I think.

dylan-chong:
Any progress on optimising this? The `module_info` spam is still causing 9 seconds of startup time for me, a bit annoying for a fast TDD cycle when tests take so long to start

PR #150, opened 2023-02-09
Move startup phase to handle_continue

luisguilhermemsalmeida:
The `init/1` function on `prometheus_sup` is doing intensive work which is delaying the initialization of applications depending on this project. This PR proposes using the `handle_continue` callback to execute functions async


luisguilhermemsalmeida:
Possible fix to https://github.com/deadtrickster/prometheus.erl/issues/81

NelsonVides:
A different fix has been implemented here: https://github.com/prometheus-erl/prometheus.erl/pull/169, in a fashion that seems to me much simpler and less heavy. Thanks for your contribution nevertheless, it was greatly indicative of where the issue may lie.

PR #154, opened 2023-07-18
Support OTP26

meox:
See: https://github.com/erlang/otp/issues/6916


NelsonVides:
This change is not needed anymore, as we now work with a different release mechanism. Thanks for your contribution nevertheless, hope to see you around again!

PR #155, opened 2023-08-20
Weighted histogram

x0id:
The `prometheus_histogram:observe_n/3,4,5` adds limited support for the weighted histograms. It accepts the extra argument "Weight" which is limited to integer numbers to fit existing bucket counters data type. It may be, for example, the number of time ticks when the recent value was observed, allowing for better accuracy when measurements are irregular.

In general, the weight could be any number, but this is difficult to implement since the overall Prometheus histogram model is based on the bucket counters concept, where the counters are expected to be non-negative integers. That's why the new function is limited to integer weights only, which is still useful in many applications.

### DIFF
--- a/src/metrics/prometheus_histogram.erl
+++ b/src/metrics/prometheus_histogram.erl
@@ -29,6 +29,13 @@
 %%                                [Method], Time).
 %%
 %% </pre>
+%%
+%% The `prometheus_histogram:observe_n/3,4,5' adds limited support for the
+%% weighted histograms. It accepts the extra argument "Weight" which is limited
+%% to integer numbers to fit existing bucket counters data type. It may be, for
+%% example, the number of time ticks when the recent value was observed,
+%% allowing for better accuracy when measurements are irregular.
+%%
 %% @end
 
 -module(prometheus_histogram).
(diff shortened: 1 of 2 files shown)


saleyn:
@deadtrickster , could you approve this one? It is a very useful patch.

deadtrickster:
surprised by Value*Weight - Value expected to be averaged by caller or observe/n called as soon as Value changes? Weight means exact measurements count? Weights are confusing to me because of neural networks and other math stuff. Could it be Count? Why not to call existing observe on each observation?

x0id:
> surprised by Value*Weight - Value expected to be averaged by caller or observe/n called as soon as Value changes?

It may be averaged, or it may be instant. My case is second one. Anyway, the value is still used to calculate the bucket. Disregarding the weight.

> Weight means exact measurements count? Weights are confusing to me because of neural networks and other math stuff. Could it be Count?

I don't care of the naming. It could be named `count`. But the `weight` is a generalization of count, sorry for math again :)

> Why not to call existing observe on each observation?

Because to keep the intention, I would have to call `observe` N times for a single observation. Remember, we deal with histogram, not the gauges. The values is mapped into the bucket number (or the bucket interval, strictly saying). And I need it to be counted in the bucket more than one time per observation.

ikvmw:
In prometheus histograms have counts, not weights.

ok, now it's more clear that you have just single event that comes with value and count. 
I'm ok merging it iff weights renamed to counts and @doc strings are very clear what value ends up in the bucket - observed or observed * count and why/

x0id:
> I'm ok merging it if weights renamed to counts and @doc strings are very clear what value ends up in the bucket - observed or observed * count and why/

Renamed weights to counts, fixed the doc.
Please let me know if anything else should be corrected.

ikvmw:
Great! I think commits must be signed

x0id:
> Great! I think commits must be signed

Is this your requirement? I've never signed my commits before. I don't even have a GPG/PGP or other signature. Why is this important?

deadtrickster:
It's a requirement for this repo, yes. I think it was made so on request of somebody wanting to use in their org. Years ago

x0id:
Commits signed.

ikvmw (approved):

deadtrickster:
❤️

PR #158, opened 2023-10-12
Move export prometheus_text_format:escape_label_value outside of TEST def

aaron-seo:
`prometheus_text_format:escape_label_value` is a useful helper function that users can benefit from using without having to duplicate the function in their code.

Namely, it can be used to escape prometheus core metric label values -- based on issue [rabbitmq/rabbitmq-server#9648](https://github.com/rabbitmq/rabbitmq-server/issues/9648) and relevant PR/review from [rabbitmq/rabbitmq-server#9656 ](https://github.com/rabbitmq/rabbitmq-server/pull/9656#discussion_r1355756241)

### DIFF
--- a/src/formats/prometheus_text_format.erl
+++ b/src/formats/prometheus_text_format.erl
@@ -77,7 +77,8 @@ format(Registry) ->
   ok = file:close(Fd),
   Str.
 
--spec escape_label_value(binary() | iolist() | undefined) -> binary().
+-spec escape_label_value(binary() | iolist()) -> binary()
+                      ; (undefined) -> no_return().
 %% @doc
 %% Escapes the backslash (\), double-quote ("), and line feed (\n) characters
 %% @end
--- a/src/formats/prometheus_text_format.erl
+++ b/src/formats/prometheus_text_format.erl
@@ -22,11 +22,12 @@
 -export([content_type/0,
          format/0,
          format/1,
-         render_labels/1]).
+         render_labels/1,
+         escape_label_value/1
+        ]).
 
 -ifdef(TEST).
 -export([escape_metric_help/1,
-         escape_label_value/1,
          emit_mf_prologue/2,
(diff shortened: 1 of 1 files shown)


src/formats/prometheus_text_format.erl
```diff
@@ -76,6 +77,15 @@ format(Registry) ->
   ok = file:close(Fd),
   Str.
 
+-spec escape_label_value(binary() | iolist() | undefined) -> binary().
```
gomoripeti, line 80: Existing type-spec is a bit strange, this function does not return `binary()` for input `undefined`

gomoripeti (approved):

src/formats/prometheus_text_format.erl
ikvmw: is it possible to fix in this PR?

deadtrickster (approved):

src/formats/prometheus_text_format.erl
aaron-seo: Looks like my local workspace didn't have GPG signatures setup correctly, so I'll create a new PR anyways with the signed commits. I'll update the spec to the following to more accurately represent the function.
```
-spec escape_label_value(binary() | iolist()) -> binary()
                      ; (undefined) -> no_return().
```

michaelklishin:
@ikvmw you may have to explicitly approve workflow runs for this PR as it is from/by a first-time contributor

aaron-seo:
May need to be merged manually even after approval, due to the job `continuous-integration/travis-ci Expected — Waiting for status to be reported` stalling. @ikvmw @deadtrickster Thanks in advance :)

aaron-seo:
Just bumping this for merge, if there are other tasks needed for the process, please feel free to let me know, thanks @deadtrickster @ikvmw

PR #169, opened 2025-02-25
Optimise startup time

### LINKED ISSUE #81: Slow Supervisor Boot Time

NelsonVides:
Fixes #81. Overrides #150.

Note too that this is already fixed in RabbitMQ for example, so it won't break its use-case: https://github.com/rabbitmq/rabbitmq-server/commit/692b6f6661b0804be14a009b0ce10da20b4ec8b2

As in the last commit:

> This means that by default prometheus won't try to find all loaded
modules implementing a behaviour, as this can be a very expensive
operation (taking many seconds as has been seen in the field). If such
behaviour was desired, it'd require explicitly setting `all_loaded` for
instrumeters and collectors. The default is the most sane behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,12 +182,12 @@ Histogram also accepts `buckets` option. Please refer to respective modules docs
 
 ### Advanced
 
-You will need these modules only if you're writing custom collector for app/lib that can't be instrumented directly.
+You will need these modules only if you're writing a custom collector for an app/lib that can't be instrumented directly.
 
 - [`prometheus_collector`](https://github.com/deadtrickster/prometheus.erl/blob/master/doc/prometheus_collector.md) - common interface for collectors;
 - [`prometheus_format`](https://github.com/deadtrickster/prometheus.erl/blob/master/doc/prometheus_format.md) - common interface for exposition formats;
 - [`prometheus_model_helpers`](https://github.com/deadtrickster/prometheus.erl/blob/master/doc/prometheus_model_helpers.md) - provides API for working with underlying Prometheus models.
-You'll use that if you want to create custom collector.
+You'll use that if you want to create a custom collector.
 
 ## Build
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -198,29 +198,27 @@ You'll use that if you want to create custom collector.
...
-prometheus_mnesia_collector,
-prometheus_summary,
-prometheus_vm_memory_collector,
-prometheus_vm_statistics_collector,
-prometheus_vm_system_info_collector
-</pre>
+- `collectors` - List of custom collectors modules to be registered automatically.
+    Can be `all_loaded` in order to find all modules implementing the `prometheus_collector` behaviour.
+    Supports an "alias" option `default`, which will append all default collectors implemented in this library.
+    If undefined, the default collectors implemented in this library will be used.
```
mikpe, line 204: So that's a behavioural change? (Not objecting, just noting.)

src/prometheus_collector.erl
```diff
@@ -128,7 +128,8 @@ Called when collector is deregistered. If collector is stateful you can put clea
 enabled_collectors() ->
     lists:usort(
         case application:get_env(prometheus, collectors) of
-            undefined -> all_known_collectors();
+            undefined -> ?DEFAULT_COLLECTORS;
+            {ok, all_loaded} -> all_known_collectors();
```
mikpe, line 132: Ok

README.md
```diff
@@ -198,29 +198,27 @@ You'll use that if you want to create custom collector.
...
+- `instrumenter` - List of custom instrumenter modules to be registered automatically.
```
mikpe: `instrumenterS` plural

src/prometheus_misc.erl
```diff
@@ -20,41 +20,35 @@
...
-    ).
+    Applications = application:loaded_applications(),
+    Modules = get_all_modules(Applications),
+    Targets = lists:usort(Modules),
+    extract_behaviour_from_modules(Targets, Behaviour).
 
+-spec get_all_modules([{atom(), string(), string()}]) -> [module()].
+get_all_modules(Applications) ->
+    AllModules = [get_modules_for_app(App) || App <- Applications],
+    lists:flatten(AllModules).
```
mikpe: Do you need deep flattening? If not `lists:flatmap/1` could simplify this.

mikpe (dismissed):

lhoguin (dismissed):
Much better than the previous behavior.

src/prometheus_sup.erl
```diff
@@ -17,10 +17,10 @@
 
 -spec start_link() -> supervisor:startlink_ret().
 start_link() ->
-    supervisor:start_link({local, ?MODULE}, ?MODULE, []).
+    supervisor:start_link({local, ?MODULE}, ?MODULE, noargs).
 
--spec init([]) -> {ok, {supervisor:sup_flags(), [supervisor:child_spec()]}}.
-init([]) ->
+-spec init(noargs) -> {ok, {supervisor:sup_flags(), [supervisor:child_spec()]}}.
+init(noargs) ->
```
lhoguin: This change is pointless.

NelsonVides: Some projects use this as a convention for several reasons, more readable tracing is my favourite. Not a big difference indeed though.

README.md
NelsonVides: Yeap. Thankfully some big projects like RabbitMQ won't break on this, but if you know of any user that might break on this point it out — check your internal use cases? :D

mikpe: I'm pretty sure we won't be affected.

src/prometheus_misc.erl
NelsonVides: Tru 👌🏽

src/prometheus_sup.erl
lhoguin: I mean, if we all sneak in our prefered styles without care like this, the project is quickly going to look weird. Please remove it, it has nothing to do with this PR.

mikpe (dismissed):

src/prometheus_sup.erl
NelsonVides: Fair enough, undone 👌🏽

NelsonVides:
@lhoguin @mikpe, just pushed a change, need another approval 🥲

mikpe (approved):

lhoguin (approved):

albertored:
Please note that this a **breaking change**.
Who is not explicitly defining collectors in application environment will see some metrics disappearing by simply updating the lib to a new **minor** version.

NelsonVides:
> Please note that this a **breaking change**. Who is not explicitly defining collectors in application environment will see some metrics disappearing by simply updating the lib to a new **minor** version.

Fair enough. So you mean this should be released as a 5.0? Can fix if that's the case.

albertored:
Ideally yes, otherwise it can be explicitly stated in the documentation and/or changelog.
It is a tricky change because if you don't have tests on all your collectors (and on the ones included from libraries) you may not discover the issue before going to production

NelsonVides:
> Ideally yes, otherwise it can be explicitly stated in the documentation and/or changelog. It is a tricky change because if you don't have tests on all your collectors (and on the ones included from libraries) you may not discover the issue before going to production

Done, released 5.0.0 and retired 4.13.0 👍🏽 

Thanks for pointing that out!

albertored:
@NelsonVides can you also release a version of prometheus_ex that allows to use the new 5.0.0?

NelsonVides:
> @NelsonVides can you also release a version of prometheus_ex that allows to use the new 5.0.0?

@albertored would love to, but actually that repo needs a revamp, like creating new CI with all the latest versions of Erlang/Elixir, a new ex_doc, fixing the configs (Mix.Config is deprecated), etc. Maybe you want to contribute a PR cleaning up that one? Would be very appreciated! 😄

albertored:
@NelsonVides https://github.com/prometheus-erl/prometheus.ex/pull/57

PR #170, opened 2025-02-25
Quantile summaries

### LINKED ISSUE #146: Reading quantile summary metrics fails intermittently when no labels are used
For the life of me I can't reproduce this in dev and uat, it only seems to fail when run in prod (where this library is getting used more heavily), and not all the time which is weird. This issue seems to occur only when labels haven't been used for quantile summary metrics.

```
** (exit) an exception was raised:
    ** (MatchError) no match of right hand side value: true
        (quantile_estimator 0.2.1) /quantile_estimator/src/quantile_estimator.erl:141: :quantile_estimator.merge/2
        (quantile_estimator 0.2.1) /quantile_estimator/src/quantile_estimator.erl:133: :quantile_estimator.compress/4
        (quantile_estimator 0.2.1) /quantile_estimator/src/quantile_estimator.erl:133: :quantile_estimator.compress/4
        (quantile_estimator 0.2.1) /quantile_estimator/src/quantile_estimator.erl:106: :quantile_estimator.compress/1
        (prometheus 4.9.1) src/metrics/prometheus_quantile_summary.erl:354: anonymous fn/3 in :prometheus_quantile_summary.collect_metrics/2
        (stdlib 3.16.1) lists.erl:1267: :lists.foldl/3
        (prometheus 4.9.1) src/metrics/prometheus_quantile_summary.erl:348: :prometheus_quantile_summary.collect_metrics/2
        (prometheus 4.9.1) src/model/prometheus_model_helpers.erl:131: :prometheus_model_helpers.create_mf/5
```

Sorry I can't be more helpful with this. I did take a look through `prometheus_quantile_summary.erl` but my inability to read Erlang code is letting me down.. A failing of mine I wish to correct at some point..

### LINKED ISSUE #159: Crashes with quantile_summary metrics and multiple schedulers
We're getting persistent crashes when emitting `quantile_summary` metrics collected from multiple schedulers:

```
** exception error: no match of right hand side value true
     in function  quantile_estimator:merge/2 (/tmp/prometheus.erl/_build/default/lib/quantile_estimator/src/quantile_estimator.erl, line 141)
     in call from quantile_estimator:compress/4 (/tmp/prometheus.erl/_build/default/lib/quantile_estimator/src/quantile_estimator.erl, line 133)
     in call from quantile_estimator:compress/4 (/tmp/prometheus.erl/_build/default/lib/quantile_estimator/src/quantile_estimator.erl, line 133)
     in call from quantile_estimator:compress/1 (/tmp/prometheus.erl/_build/default/lib/quantile_estimator/src/quantile_estimator.erl, line 106)
     in call from prometheus_quantile_summary:'-values/2-fun-0-'/3 (/tmp/prometheus.erl/src/metrics/prometheus_quantile_summary.erl, line 309)
     in call from lists:foldl_1/3 (lists.erl, line 1355)
     in call from prometheus_quantile_summary:values/2 (/tmp/prometheus.erl/src/metrics/prometheus_quantile_summary.erl, line 306)
```

There appears to be two contributing causes:
1. `quantile_estimator:compress/1` and its helper function `merge/2` want to assert that the input is as expected (i.e., as `quantile_estimator` itself would have produced), but the assertion on line 141 fails.
2. `prometheus_quantile_summary` records observations from different schedulers under different keys in its ETS table (to reduce contention). Then it wants to combine observations with `quantile_merge/2`, which just appends the observations from the different schedulers (the `++` on line 497), before passing that to `quantile_estimator:compress/1`.

The problem, as far as I understand this code, is that the append in `quantile_merge/2` is not guaranteed to result in data that `quantile_estimator:merge/2` considers to be well-formed.

I'm attaching a test case (zipped due to github limitations). It's based on the actual contents of the `prometheus_quantile_summary_table` ETS from one of our crashes, reduced to the bare minimum that still reproduced the crash. (I've tried to come up with a test case that only used the public APIs, but failed, possibly due to the non-determinism described below.)
[bug.erl.zip](https://github.com/deadtrickster/prometheus.erl/files/13114723/bug.erl.zip)

Note: `quantile_summary_metrics.erl` contains two non-deterministic constructs (retrieving multiple entries from a `set` ETS table) and I found it necessary to eliminate that non-determinism in order to have a reliable test case:

```
diff --git a/src/metrics/prometheus_quantile_summary.erl b/src/metrics/prometheus_quantile_summary.erl
index 9ffdd3e..e860bb0 100644
--- a/src/metrics/prometheus_quantile_summary.erl
+++ b/src/metrics/prometheus_quantile_summary.erl
@@ -290,7 +290,7 @@ value(Registry, Name, LabelValues) ->
                             [],
                             ['$$']}]) of
     [] -> undefined;
-    Values -> {Count, Sum, QE} = reduce_values(Values),
+    Values -> {Count, Sum, QE} = reduce_values(lists:sort(Values)),
               {Count,  prometheus_time:maybe_convert_to_du(DU, Sum), quantile_values(QE, QNs)}
   end.
 
@@ -403,7 +403,7 @@ insert_metric(Registry, Name, LabelValues, Value, ConflictCB) ->
   end.
 
 load_all_values(Registry, Name) ->
-  ets:match(?TABLE, {{Registry, Name, '$1', '_'}, '$2', '$3', '$4', '_'}).
+  lists:sort(ets:match(?TABLE, {{Registry, Name, '$1', '_'}, '$2', '$3', '$4', '_'})).
 
 get_configuration(Registry, Name) ->
   MF = prometheus_metric:check_mf_exists(?TABLE, Registry, Name),
```
[0001-eliminate-non-deterministic-behaviour.patch.txt](https://github.com/deadtrickster/prometheus.erl/files/13114767/0001-eliminate-non-deterministic-behaviour.patch.txt)

This is with `prometheus.erl` `v4.10.0`, any recent OTP (24.3.4.14, 25.3.2.7, 26.1.2), and Linux/x86_64 (AL2, Fedora 38).

NelsonVides:
Fixes #159 and #146 

As in the last commit:

> Note that the underlying algorithm for quantiles is not mergeable,
hence, it requires a linear strategy. However, precisely because it is
not generally mergeable, and it is also expensive, Prometheus generally
recommends using histograms on the client side, and it is left to the
Prometheus server to compute, merge, and aggregate, histograms,
recreating any required quantiles.
> It is left for the future to implement mergeable summaries, an idea
would be a DDSketch algorithm or any variant thereof.

### DIFF
--- a/src/metrics/prometheus_quantile_summary.erl
+++ b/src/metrics/prometheus_quantile_summary.erl
@@ -108,7 +108,6 @@ and histograms are recommended instead.
 -define(SUM_POS, 3).
 -define(COUNTER_POS, 2).
 -define(QUANTILE_POS, 4).
--define(WIDTH, 16).
 
 ?DOC("""
 Creates a summary using `Spec`.
@@ -267,9 +266,8 @@ remove(Registry, Name, LabelValues) ->
         lists:flatten([
             ets:take(
                 ?TABLE,
-                {Registry, Name, LabelValues, Scheduler}
+                {Registry, Name, LabelValues}
             )
-         || Scheduler <- schedulers_seq()
         ])
     of
         [] -> false;
@@ -302,10 +300,9 @@ reset(Registry, Name, LabelValues) ->
         lists:usort([
             ets:update_element(
                 ?TABLE,
-                {Registry, Name, LabelValues, Scheduler},
+                {Registry, Name, LabelValues},
                 [{?COUNTER_POS, 0}, {?SUM_POS, 0}, {?QUANTILE_POS, new_quantile(Configuration)}]
             )
-         || Scheduler <- schedulers_seq()
         ])
     of
         [_, _] -> true;
@@ -341,7 +338,7 @@ value(Registry, Name, LabelValues) ->
     MF = prometheus_metric:check_mf_exists(?TABLE, Registry, Name, LabelValues),
     DU = prometheus_metric:mf_duration_unit(MF),
     #{quantiles := QNs} = prometheus_metric:mf_data(MF),
-    Spec = [{{{Registry, Name, LabelValues, '_'}, '$1', '$2', '$3', '_'}, [], ['$$']}],
+    Spec = [{{{Registry, Name, LabelValues}, '$1', '$2', '$3', '_'}, [], ['$$']}],
     case ets:select(?TABLE, Spec) of
         [] ->
             undefined;
@@ -399,7 +396,7 @@ values(Registry, Name) ->
 -spec deregister_cleanup(prometheus_registry:registry()) -> ok.
 deregister_cleanup(Registry) ->
     prometheus_metric:deregister_mf(?TABLE, Registry),
-    true = ets:match_delete(?TABLE, {{Registry, '_', '_', '_'}, '_', '_', '_', '_'}),
+    true = ets:match_delete(?TABLE, {{Registry, '_', '_'}, '_', '_', '_', '_'}),
     ok.
 
 ?DOC(false).
@@ -445,7 +442,7 @@ collect_metrics(Name, {CLabels, Labels, Registry, DU, Configuration}) ->
 %%====================================================================
 
 deregister_select(Registry, Name) ->
-    [{{{Registry, Name, '_', '_'}, '_', '_', '_', '_'}, [], [true]}].
+    [{{{Registry, Name, '_'}, '_', '_', '_', '_'}, [], [true]}].
 
 validate_summary_spec(Spec) ->
     Labels = prometheus_metric_spec:labels(Spec),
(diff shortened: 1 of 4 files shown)


mikpe (dismissed):

rebar.config
```diff
@@ -19,7 +19,7 @@
 ]}.
 
 {deps, [
-    {quantile_estimator, "1.0.1"}
+    {quantile_estimator, "~> 1.0"}
```
lhoguin: Please use specific versions for now as this will break current RabbitMQ. Erlang.mk doesn't autopatch fuzzy versions properly yet. I will open a ticket in Erlang.mk to do it at some point.

lhoguin (changes requested):

rebar.config
NelsonVides: Good to know! Will update and ping you, thank you.

DenysGonchar (dismissed):

rebar.config
NelsonVides: Fixed now 👌🏽

lhoguin (approved):
This review only validates that RabbitMQ should continue to work after this change. I have not evaluated the technical merits of quantile summaries (I'm out of my league there).

mikpe (approved):

PR #173, opened 2025-03-03
Optimisations and cleanups

NelsonVides:
This PR proposes a miscellaneous of updates and improvements to the repo. Commits are well organised by topics and with good descriptions, so that'd be the best way to review the whole thing, as some changes are not related to each other – but are some times too small to be moved to their own PR, which I could easily do if any of you would prefer me to do anyways.

---
EDIT: https://github.com/prometheus-erl/prometheus.erl/pull/173#issuecomment-2701199656

### DIFF
--- a/src/model/prometheus_model_helpers.erl
+++ b/src/model/prometheus_model_helpers.erl
@@ -192,18 +192,15 @@ boolean_metric(Labels, Value0) ->
 -spec boolean_value(Value) -> RealValue when
     Value :: prometheus:prometheus_boolean(),
     RealValue :: undefined | 0 | 1.
-boolean_value(Value) ->
-    case Value of
-        true -> 1;
-        false -> 0;
-        1 -> 1;
-        0 -> 0;
-        [] -> 0;
-        _ when is_number(Value) andalso Value > 0 -> 1;
-        _ when is_list(Value) -> 1;
-        undefined -> undefined;
-        _ -> erlang:error({invalid_value, Value, "value is not boolean"})
-    end.
+boolean_value(true) -> 1;
+boolean_value(false) -> 0;
+boolean_value(undefined) -> undefined;
+boolean_value(1) -> 1;
+boolean_value(0) -> 0;
+boolean_value([]) -> 0;
+boolean_value([_ | _]) -> 1;
+boolean_value(Value) when is_number(Value), Value > 0 -> 1;
+boolean_value(Value) -> erlang:error({invalid_value, Value, "value is not boolean"}).
 
 ?DOC(#{equiv => lists:map(fun counter_metric/1, Values)}).
 -spec counter_metrics(Specs) -> [prometheus_model:'Metric'()] when
--- a/src/prometheus_buckets.erl
+++ b/src/prometheus_buckets.erl
@@ -131,14 +131,8 @@ linear(Start, Step, Count) ->
     linear(Start, Step, Count, []).
 
 -spec position(buckets(), number()) -> pos_integer().
-position(Buckets, Value) ->
-    position(
-        Buckets,
-        fun(Bound) ->
-            Value =< Bound
-        end,
-        0
-    ).
+position(Buckets, Value) when is_list(Buckets), is_number(Value) ->
+    find_position(Buckets, Value, 0).
 
 linear(_Current, _Step, 0, Acc) ->
     lists:reverse(Acc);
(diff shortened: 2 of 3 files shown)


src/metrics/prometheus_counter.erl
```diff
@@ -90,7 +90,7 @@ inc(Caller) ->
 -define(TABLE, ?PROMETHEUS_COUNTER_TABLE).
 -define(ISUM_POS, 2).
 -define(FSUM_POS, 3).
--define(WIDTH, 16).
+-define(WIDTH, erlang:system_info(schedulers_online)).
```
onno-vos-dev: Maybe instead of fixing this to be `schedulers_online` we can set this up as an `application:get_env(<application>, <some_name>, 16)` so the original remains preserved but we open up the loophole for setting a higher value.

src/metrics/prometheus_histogram.erl
```diff
@@ -95,7 +95,7 @@ the number of time ticks when the recent value was observed).
 -define(ISUM_POS, 3).
 -define(FSUM_POS, 4).
 -define(BUCKETS_START, 5).
--define(WIDTH, 16).
+-define(WIDTH, erlang:system_info(schedulers_online)).
```
onno-vos-dev: Maybe instead of fixing this to be `schedulers_online` we can set this up as an `application:get_env(<application>, <some_name>, 16)` so the original remains preserved but we open up the loophole for setting a higher value.

onno-vos-dev (commented):
Haven't had the time to look through all commits yet but sharing these two review comments as a start

src/metrics/prometheus_counter.erl
mikpe: Another issue is that `schedulers_online` can change value dynamically at runtime.

NelsonVides: Right. Should use just `schedulers`, as that can't change at runtime.

On the other hand I'm not sure it'd help to make it configurable, currently the algorithm that chooses the key would do `X band (?WIDTH - 1)`, so most values would mean skewing the distribution of schedulers to each key and some keys would see more contention than others. Sensibly, if we're keying on the scheduler id, the only values that make sense are multiples (or halves) of the number of schedulers.

NelsonVides:
I guess this PR is too big to review, so I've split it into #175 for the uncontroversial, #176 for the most convoluted, and left here maybe the most interesting changes.

lhoguin:
About the change to use number of schedulers for the width, what if I have a very large number of schedulers, will it still work? Or is there a limit in ets?

src/metrics/prometheus_histogram.erl
DenysGonchar: This solution has the same issue as @mikpe mentioned above, we don't want to let users change that value dynamically. another issue with defaults - they are rarely changed :)

src/metrics/prometheus_counter.erl
```diff
@@ -90,7 +90,7 @@ inc(Caller) ->
 -define(TABLE, ?PROMETHEUS_COUNTER_TABLE).
 -define(ISUM_POS, 2).
 -define(FSUM_POS, 3).
--define(WIDTH, 16).
+-define(WIDTH, erlang:system_info(schedulers)).
```
DenysGonchar: `X band (?WIDTH - 1)` - it seems that WIDTH is expected to be a power of 2

src/metrics/prometheus_histogram.erl
```diff
@@ -95,7 +95,7 @@ the number of time ticks when the recent value was observed).
 -define(ISUM_POS, 3).
 -define(FSUM_POS, 4).
 -define(BUCKETS_START, 5).
--define(WIDTH, 16).
+-define(WIDTH, erlang:system_info(schedulers)).
```
DenysGonchar: `X band (?WIDTH - 1)` - it seems that WIDTH is expected to be a power of 2

NelsonVides:
On @lhoguin comment, I'd say bigger width should just make reads slower as there'd be more rows to read, but inserts should be faster as there'd be less contention. On the other hand, these are educated guesses, not measured points.

On @DenysGonchar comment, actually that's a **very** good point. In general, what we want is to ensure the distribution of values is as uniform as possible across schedulers, I just ran some checks and it seems like indeed my proposal wouldn't be good enough either.

For that matter, I'm just taking this change out of this PR, will later on when I find time for it run some actual statistics about distributions, and benchmark different solutions. The rest of the changes still apply.

Thank you for your comments 🙇🏽

DenysGonchar (approved):
looks good to me.

PR #174, opened 2025-03-05
Fix rebar.config.script when the src dir doesn't exist

NelsonVides:
This can happen when compiling a project that uses this library as a dependency and has no `src/` directory, arguably an edge case though.

The intend of the script is to prepare sections in ex_doc according to the directories, so that metric modules are grouped into a metrics section, etc.


mikpe (approved):

DenysGonchar (approved):
👌

PR #176, opened 2025-03-05
Optimise handling of integers and floats for counters, gauges, summaries

NelsonVides:
The point is that these tables keep separate positions in the tuple for
integers and for floats, as on the integer position we can use the
optimised `ets:update_counter/3,4`, and probabilistically integers are
the most common input.

The issue is that under some circumstances, integer values would be
appended to the float position, which, while correct, misses the
opportunity for the optimisation.

Note too that gauges could take the value `undefined` to effectively
block changes until explicitly set. This was implemented, but not
tested, and in some edge-cases (the integer/float optimisation) it would
be buggy.

Split from https://github.com/prometheus-erl/prometheus.erl/pull/173

### DIFF
--- a/src/metrics/prometheus_counter.erl
+++ b/src/metrics/prometheus_counter.erl
@@ -87,6 +87,10 @@ inc(Caller) ->
 -behaviour(prometheus_metric).
 -behaviour(prometheus_collector).
 
+%% Records will be triplets `{Key, ISum, FSum}`
+%%  where `ISum` are integer sums, and `FSum` are float sums
+%% Keys use the scheduler index, because counters by definition aggregate
+%%  even after parallel updates. Hence we can run updates without locks.
 -define(TABLE, ?PROMETHEUS_COUNTER_TABLE).
 -define(ISUM_POS, 2).
 -define(FSUM_POS, 3).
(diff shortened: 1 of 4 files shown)


src/metrics/prometheus_counter.erl
```diff
@@ -349,9 +353,16 @@ collect_metrics(Name, {CLabels, Labels, Registry}) ->
 deregister_select(Registry, Name) ->
     [{{{Registry, Name, '_', '_'}, '_', '_'}, [], [true]}].
 
-insert_metric(Registry, Name, LabelValues, Value, ConflictCB) ->
-    prometheus_metric:check_mf_exists(?TABLE, Registry, Name, LabelValues),
+insert_metric_int(Registry, Name, LabelValues, Value, ConflictCB) ->
```
onno-vos-dev, line 356: Not that great in compiling beam code in my brain but wouldn't it help to hint the compiler with a `is_integer(Value)` guard here or is it sufficient enough to have a separate function for it? 🤔 

(Same all the other specific functions)

NelsonVides: I'd say this function is _only_ called from another function that already did the `is_integer(Value)` guard, so I'm pretty sure the compiler already generated optimised code for it. Haven't verify it though 🤔

NelsonVides:
@prometheus-erl/prometheus-erl any reviewer here? Just planning a new release after this minor optimisation is merged :)

onno-vos-dev:
> @prometheus-erl/prometheus-erl any reviewer here? Just planning a new release after this minor optimisation is merged :)

Sorry my bad! Never pressed approve after your last reply!

onno-vos-dev (approved):

PR #177, opened 2025-03-24
Summary sketch

NelsonVides:
This uses now [ddskerl](https://hex.pm/packages/ddskerl), an implementation of the DDSketch algorithm, instead of [quantile_estimator](https://hex.pm/packages/quantile_estimator), which is in turn an implementation of Cormode's biased algorithms.

The most important difference is that the biased algorithms aren't mathematically mergeable, hence keeping a per-scheduler summary would never work, and keeping a single global one would never scale (and not mentioning race conditions on inserts as the data-structure is not possible to treat atomically in an ets table). Instead, the DDSketch algorithm is fully mergeable, and the data structure implementation has an ETS backend fully tested for all sorts of race conditions.

As a possibility that is out of the scope of this PR, a custom exporter might provide the Sketch data structure to datadog, so that the quantile merges can be done on the server, and that way the metrics server can aggregate quantiles across many hosts in a meaningful way.


lhoguin:
What problem are you trying to solve in this PR?

NelsonVides:
> What problem are you trying to solve in this PR?

Having quantile summaries entirely. The previous ones were known to be broken (see https://github.com/prometheus-erl/prometheus.erl/pull/170, https://github.com/prometheus-erl/prometheus.erl/issues/146, https://github.com/prometheus-erl/prometheus.erl/issues/159). My fix at https://github.com/prometheus-erl/prometheus.erl/pull/170 was just to stop it from breaking, but they were still not correct due to data-loss during race conditions, and surely not performant as they were doing lots of lookup-inserts of big data structures, when this PR does only `update_counters`.

rebar.config
```diff
@@ -35,13 +35,14 @@
 
 {dialyzer, [
     {warnings, [no_return, error_handling, unknown]},
-    {plt_extra_apps, [mnesia, quantile_estimator]}
+    {plt_extra_apps, [mnesia, ddskerl]}
```
paulo-ferraz-oliveira: Would a better approach be to add it to `.app.src`?

src/prometheus_time.erl
```diff
@@ -155,14 +155,13 @@ maybe_convert_to_native(DU, Value) ->
         _ -> to_native(Value, DU)
     end.
 
--spec maybe_convert_to_du(duration_unit(), infinity | number()) -> infinity | number().
+-spec maybe_convert_to_du(undefined | duration_unit(), infinity | number()) -> infinity | number().
+maybe_convert_to_du(undefined, Value) ->
+    Value;
 maybe_convert_to_du(_, infinity) ->
     infinity;
```
paulo-ferraz-oliveira: Did (`undefined`, `infinity`) ever make sense before? Because we're losing that with this change.

rebar.config
```diff
@@ -19,7 +19,7 @@
 ]}.
 
 {deps, [
-    {quantile_estimator, "1.0.2"}
+    {ddskerl, "0.2.1"}
```
paulo-ferraz-oliveira: Is `ddskerl` battle-tested? I saw it mentioned in the Erlang forums, but maybe it could be an "alternative" in this library and when it's fully battle-tested made the default?

src/metrics/prometheus_quantile_summary.erl
```diff
@@ -45,33 +45,18 @@ request_size_bytes\{quantile=\"0.95\"\}
...
-### Notes
-
-The underlying algorithm implemented in `m:quantile_estimator` does not support mergeability of
-summaries, therefore a summary in this implementation is linear, that is, all updates happen on the
-same ets record. Because of this, race conditions can imply datapoint loss.
-Statistically, this might not be relevant, but such impact is not ensured.
-
-Because of this, support for quantile summaries is rather experimental
-and histograms are recommended instead.
+See `t:ddskerl_ets:opts/0` for configuration options. These need to be passed as a proplist.
```
paulo-ferraz-oliveira, line 48: Doesn't this mean that a wrapper function for this would require importing `ddskerl` for Dialyzer to not complain? Would a better approach be to replicate the type with a different name in `prometheus.erl`?

paulo-ferraz-oliveira: (unless this is not supposed to be exported interface, ofc)

rebar.config
lhoguin: The point of the PR is to change the algorithm, and the algorithm is implemented in quantile_estimator/ddskerl. It's not just switching a dependency for another.

Perhaps we can require Nelson to run this PR in production for a while before merging if you're worried about the scale of the change.

NelsonVides: They're not equivalent, but `quantile_estimator` wasn't battle-tested before it was added: the algorithm is not mergeable and it uses a data-structure that is not paralellisable either.

The DDSketch algorithm is surely battle-tested, by DataDog themselves: https://www.datadoghq.com/blog/engineering/computing-accurate-percentiles-with-ddsketch/. `ddskerl` is the new Erlang port (I'm the author btw 🙂). Ideally it can also allow to implement an exporter using protobufs as in https://github.com/DataDog/sketches-js/blob/master/src/ddsketch/proto/DDSketch.proto when integrating with DataDog, so that the exported data can still be mergeable on the server, but that surely falls outside of the scope of this repo.

Whether this branch with this dependency specifically is battle-tested, well, not yet in prod. I'm using it locally and already pushing it to some staging environments. My original idea is to push this to master but not release a Hex package yet, just pull the dep as a git from master, but same thing can happen if pulling from this branch, just thought master was more discoverable. I can add a note to the PR for people to see and choose if they want to try it out and come back to you in a few weeks when it is prod-tested though 👌🏽

NelsonVides:
I've pushed a couple of changes to the default quantiles, in order not create any backwards incompatible change. As such, previously created quantiles using `targets` and `compres_limit` keys would not fail, but simply such keys would be ignored.

src/metrics/prometheus_quantile_summary.erl
NelsonVides: It is not an exported interface, only documentation. Specs require here to take proplists.

src/prometheus_time.erl
NelsonVides: Very good question. Just checking, coverage had actually never hit this clause. This is called from gauges, histograms, summaries, and quantile_summaries. Quantile summaries is the work in this PR, and it cannot be `infinity`. Summaries are doing number addition with `+` before giving in to this helper, so it cannot be an atom either. Histograms are doing `+` or `lists:sum/1`, so again must be integers. Interestingly, only gauges can actually pass an atom as the second argument, and they'll pass `undefined`, which would crash, but that wasn't tested. Adding a test for that.

Also, `(undefined, infinity)` would match the second clause, which just as before, would return infinity, so no real functional change is lost.

paulo-ferraz-oliveira: > (undefined, infinity) would match the second clause

You mean in the past? Because as is (check above) it wouldn't, right? As soon as it's `undefined` you get clause `(undefined, Value)`.

In any case, I'm Ok with the change and since you're gonna put some tests on, even better.

rebar.config
paulo-ferraz-oliveira: > The DDSketch algorithm is surely battle-tested

I didn't mean the algo, but the implementation 😄 I'm sure it's Ok, I just wanted to check what steps had been taken previously to ensure said Ok-ness (as @lhoguin points out: "if you're worried about the scale of the change.")

> I can add a note to the PR for people to see and choose if they want to try it out and come back to you in a few weeks when it is prod-tested though 👌🏽

This works for me.

rebar.config
```diff
@@ -126,7 +127,8 @@
                 ]
             }},
             {elvis_style, god_modules, #{limit => 54}},
-            {elvis_style, dont_repeat_yourself, #{min_complexity => 50}}
+            {elvis_style, dont_repeat_yourself, #{min_complexity => 50}},
+            {elvis_style, no_catch_expressions, disable}
```
paulo-ferraz-oliveira, line 134: Are we using `catch` expressions, though? If I remember correctly there's measurable overhead when creating the stacktrace to not use it, but:

1. maybe it's better in OTP these days?
2. I don't know all the use cases where we have `catch`

rebar.config
NelsonVides: Done 👌🏽

src/prometheus_time.erl
NelsonVides: I mean, as soon as the first argument is undefined, it'll match the first clause, and return `Value`, which, if the function was called with `(undefined, infinity)`, it would return `infinity`, which is the same effect that `(undefined, infinity)` was having before, so no change of behaviour there 😄 

In any case, see the new tests in the recent commits :)

rebar.config
NelsonVides: Here https://erlangforums.com/t/otp-28-0-rc1-released/4482/10 Bjorn said that
> If the value of the catch is ignored:
> ```
> _ = catch Expr
> ```
> the compiler internally rewrites the catch to a try…catch, which does not need to build the stacktrace.

So performance-wise there's no impact. The catch is being used only in that same pattern, when reading the value and looping over the width in order to ignore if any wasn't used yet and would fail to read.

paulo-ferraz-oliveira: Oh, that sounds like a new one, and a great one at that. Might just have to suggest something next to Elvis Core for this update. Thanks.

src/prometheus_time.erl
paulo-ferraz-oliveira: True. Good stuff.

NelsonVides:
@paulo-ferraz-oliveira @lhoguin so, this has been in prod for a month already and it's working wonderfully 🔥

Maybe time to review and get it merged?

PS: for the record, the current history is a mess because as there might be applications depending on the branch, force-pushing might break builds when the hash disappeared, so I've had to keep the history untouched. When merging, I'll need to squash-merge, and then ensure this branch doesn't disappear for at least a few months :)

@mkuratczyk, when we merge this, we can release a new version together with all the latest versions of maps you've contributed recently too. v6 I suppose, as this is a breaking change? 🤔

NelsonVides:
Hello there, any available reviews here? 🙏🏽

mkuratczyk:
maps support was not breaking (I hope :) ). I can try to review this PR in a few hours

NelsonVides:
> maps support was not breaking (I hope :) ). I can try to review this PR in a few hours

Well, probably. Going to review what we had so far and if I don't see anything breaking then will release 5.1.0 👌🏽

Thanks! :)

paulo-ferraz-oliveira (dismissed):

paulo-ferraz-oliveira:
I was keeping tabs on what was changing, and nothing surprised me, so I'm approving.

mkuratczyk (dismissed):

NelsonVides:
Thanks for the approvals! Merged the latest master and fixed a documentation typo, will bypass stale reviews and merge, hopefully nobody has a problem with that :)

PR #180, opened 2025-05-20
Allow maps of metrics in create_mf/4

mkuratczyk:
```
create_mf(<<"foo">>, "help", gauge, #{#{<<"l1">> => <<"v1">>} => my_value})
```

is equivalent with:
```
create_mf(<<"foo">>, "help", gauge, [{[{<<"l1">>, <<"v1">>}], my_value}])
```

### DIFF
--- a/src/model/prometheus_model_helpers.erl
+++ b/src/model/prometheus_model_helpers.erl
@@ -94,9 +94,12 @@ Create Metric Family of `Type`, `Name` and `Help`.
     Type :: atom(),
     Metrics ::
         [prometheus_model:'Metric'()]
+        | #{#{prometheus:label_name() => prometheus:label_value()} => prometheus:value()}
         | prometheus_model:'Metric'()
         | prometheus:metrics(),
     MetricFamily :: prometheus_model:'MetricFamily'().
+create_mf(Name, Help, Type, Metrics) when is_map(Metrics) ->
+    create_mf(Name, Help, Type, maps:to_list(Metrics));
 create_mf(Name, Help, Type, Metrics0) ->
     Metrics = metrics_from_tuples(Type, Metrics0),
     #'MetricFamily'{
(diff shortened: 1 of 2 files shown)


NelsonVides:
Noticed this, seems nice but still want to ask, what's the use-case? Where is this used? 🤔

mkuratczyk:
We have a [module for tracking counters](https://github.com/rabbitmq/seshat), which is likely used only by RabbitMQ. Everything in seshat is map-based, so when we want to export Seshat counters as Prometheus metrics, we need to do some conversions. Given that prometheus.erl already does a lot of type conversions (and that maps are better than proplists, according to me ;) ), it feels convenient to allow maps as a proplist alternative.

NelsonVides:
I 100% support maps>>>proplists and I wish prometheus itself didn't have so many proplist types hardcoded everywhere 😛 

In any case, is your exporter to prometheus counters module public? Would love to have a look at how this is used, but also maybe get an idea on how to do these things myself :)

mkuratczyk:
it is public, but "it depends" and I doubt you can learn much Prometheus-wise there. There's a "built-in" Prometheus serializer on this branch: https://github.com/rabbitmq/seshat/blob/prometheus-support/src/seshat.erl#L303-L356
But after implementing it, I decided to leave it in (in theory, there could be other Seshat users who'll find it useful), but actually on my RabbitMQ branch where I'm working right now, I do this: https://github.com/rabbitmq/rabbitmq-server/blob/expose-ra-counters/deps/rabbitmq_prometheus/src/collectors/prometheus_rabbitmq_raft_metrics_collector.erl#L31-L46

(this already relies on this PR)

NelsonVides:
Man, the second option is just sooo much cleaner. That's the template I imagined you'd be using, nice to see it done.

NelsonVides (approved):

PR #184, opened 2025-05-25
Upgrade OTP versions, supporting now only 26/27/28 and adapting to enhancements

NelsonVides:
Takes advantage of `dynamic()` typing as well as of `ets:lookup_element/4`.

Tagging everyone as it might be a more "controversial" change.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,11 +11,11 @@ on:
 
 jobs:
   test:
-    name: OTP ${{matrix.otp}}
+    name: OTP ${{matrix.otp}} - rebar3 ${{matrix.rebar3}}
     strategy:
       matrix:
         otp: ['28', '27', '26']
-        rebar3: ['3.24']
+        rebar3: ['3.24', '3.25']
     runs-on: 'ubuntu-24.04'
     env:
       OTPVER: ${{ matrix.otp }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,8 +14,8 @@ jobs:
     name: OTP ${{matrix.otp}}
     strategy:
       matrix:
-        otp: ['27', '26', '25']
-        rebar3: ['3.24.0']
+        otp: ['28', '27', '26']
+        rebar3: ['3.24']
     runs-on: 'ubuntu-24.04'
     env:
       OTPVER: ${{ matrix.otp }}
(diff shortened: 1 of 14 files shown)


mkuratczyk (dismissed):

mikpe (dismissed):

.github/workflows/main.yml
```diff
@@ -14,8 +14,8 @@ jobs:
     name: OTP ${{matrix.otp}}
     strategy:
       matrix:
-        otp: ['27', '26', '25']
-        rebar3: ['3.24.0']
+        otp: ['28', '27', '26']
+        rebar3: ['3.24']
```
paulo-ferraz-oliveira: Maybe we could already target `3.25`, out a few days ago?

src/metrics/prometheus_histogram.erl
```diff
@@ -241,25 +241,23 @@ Raises:
     Count :: integer().
 observe_n(Registry, Name, LabelValues, Value, Count) when is_integer(Value), is_integer(Count) ->
     Key = key(Registry, Name, LabelValues),
-    try ets:lookup_element(?TABLE, Key, ?BOUNDS_POS) of
-        Buckets ->
+    case ets:lookup_element(?TABLE, Key, ?BOUNDS_POS, undefined) of
+        undefined ->
+            insert_metric(Registry, Name, LabelValues, Value, Count, fun observe_n/5);
```
paulo-ferraz-oliveira, line 246: This seems like a fix, not adapting to a latest OTP. Where did this surface from?

src/prometheus_metric.erl
```diff
@@ -75,7 +75,7 @@ as well as handling metric labels and data.
         registry => prometheus_registry:registry(),
         constant_labels => [{atom(), term()}],
         labels => labels(),
-        data => any(),
+        data => term(),
```
paulo-ferraz-oliveira, line 78: If we wanna be consistent, we can introduce `rebar3_lint` rule `consistent_generic_type`. I'm Ok to introduce `rebar3_lint` with a very limited ruleset, if you want... but that's another issue. I just noticed `any()` was being changed to `term()` and that triggered this thought in me.

paulo-ferraz-oliveira (dismissed):
Left comments, but overall the change is acceptable, for me.

.github/workflows/main.yml
NelsonVides: Oh, we got 3.25 already, didn't notice, awesome!

src/metrics/prometheus_histogram.erl
NelsonVides: This one's taking advantage of a OTP26 feature, `ets:lookup_element/3` would raise if the element didn't exist. `ets:lookup_element/4` was introduced here https://github.com/erlang/otp/commit/e528e26bf76267db4ff450d598c9a816d8d4a4ce and it's just a matter of being more ergonomic and also faster, no need to throw-catch.

src/prometheus_metric.erl
NelsonVides: Didn't know that rule, going to have a look at it 👌🏽

src/metrics/prometheus_histogram.erl
paulo-ferraz-oliveira: Cool stuff. I guess the title of the PR is a bit misleading then. 😄 You're not "upgrading to support", you're "adapting to recent enhancements" 👍 good stuff, in any case.

src/prometheus_metric.erl
paulo-ferraz-oliveira: The rule actually proposes what you did, and is proposed by Erlang/OTP: `term()` over `any()`, since more idiomatic.

NelsonVides:
Anybody wants to reapprove? Previous approvals were restarted after the last pushed changes :)

mkuratczyk (approved):

PR #185, opened 2025-06-21
Upgrade plugins and fix new linter errors

NelsonVides:
@paulo-ferraz-oliveira tagging you as I know you were working on the last release of Elvis :)

### DIFF
--- a/src/collectors/vm/prometheus_vm_system_info_collector.erl
+++ b/src/collectors/vm/prometheus_vm_system_info_collector.erl
@@ -222,12 +222,12 @@ metric_enabled(Name, Metrics) ->
 
 collect_allocator_metrics() ->
     Metrics = lists:flatten([
-        collect_allocator_metrics_(Alloc, Instance, Info)
+        collect_allocator_metrics_1(Alloc, Instance, Info)
      || {{Alloc, Instance}, Info} <- allocators()
     ]),
     prometheus_model_helpers:gauge_metrics(Metrics).
 
-collect_allocator_metrics_(Alloc, Instance, Info) ->
+collect_allocator_metrics_1(Alloc, Instance, Info) ->
     [
         [
             allocator_metric(Alloc, Instance, Kind, Key, KindInfo)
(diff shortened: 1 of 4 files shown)


paulo-ferraz-oliveira:
Regarding

```erlang
    [
        {{A, N}, lists:sort(proplists:delete(versions, Props))}
     || A <- Allocators,
        Allocs <- [erlang:system_info({allocator, A})],
        Allocs =/= false,
        {_, N, Props} <- Allocs
    ].
```

I haven't tested this, but 

```erlang
    [
        {{A, N}, lists:sort(proplists:delete(versions, Props))}
     || A <- Allocators,
        {_, N, Props} <- [erlang:system_info({allocator, A})]
    ].
```

should work, even if it's less readable (because you lose the filter, which is implicit).

paulo-ferraz-oliveira:
Avoiding rules is also a possibility, but I tend not to do it at the project level, but start at the function level. Instead of `disable` you can always use `ignore => [{M, F, A}]`.

NelsonVides:
Ok, went around what the lists are supposed to be and found a different fix, it should be good to go now. Review plis? 🙏🏽😄

src/collectors/vm/prometheus_vm_system_info_collector.erl
```diff
@@ -255,8 +255,9 @@ allocators() ->
     [
         {{A, N}, lists:sort(proplists:delete(versions, Props))}
      || A <- Allocators,
-        Allocs <- [erlang:system_info({allocator, A})],
-        Allocs =/= false,
```
paulo-ferraz-oliveira, line 259: Hurray. That's what the rule is there for, to make us think about alternatives. If there's none we avoid it, but in this case there was a good solution for us 😄

paulo-ferraz-oliveira (dismissed):

src/collectors/vm/prometheus_vm_system_info_collector.erl
NelsonVides: Actually, I just changed the code to make it more explicit, I find this list comprehension quite a terse hack, and also, what if SMP is disabled for ERTS (why tf would do that but you never know), then the number of allocators would be actually just one and this would return empty. Better make it just explicit, wdyt? :)

paulo-ferraz-oliveira: Oh, you changed it. Lemme give it a quick look.

paulo-ferraz-oliveira: I like explicit, yeah; if it works, it works. I just though you were going for minimal change, so the previous one (even if implicit) also worked for me.

paulo-ferraz-oliveira (approved):

PR #186, opened 2025-06-22
Use write_concurrency=auto for better scalability

NelsonVides:
Benchmarks generally show better scalability as the parallelism increases, on my 12CPU hardware it is already visible:

image

`auto` is available since OTP25, so we're good to go version-wise.


DenysGonchar (approved):
looks good

PR #188, opened 2025-07-15
Run promtool in CI

### LINKED ISSUE #135: Built-in collectors fail promtool linting
With a stock install of `prometheus.ex` exporting the default Erlang metrics, the following linting failures are reported:

```
$ wget -O - -q http://localhost:9548/metrics | promtool check metrics
erlang_vm_statistics_context_switches counter metrics should have "_total" suffix
erlang_vm_statistics_garbage_collection_bytes_reclaimed counter metrics should have "_total" suffix
erlang_vm_memory_atom_bytes_total non-counter metrics should not have "_total" suffix
erlang_vm_statistics_garbage_collection_number_of_gcs counter metrics should have "_total" suffix
erlang_vm_statistics_garbage_collection_words_reclaimed counter metrics should have "_total" suffix
erlang_vm_statistics_runtime_milliseconds counter metrics should have "_total" suffix
erlang_vm_statistics_wallclock_time_milliseconds counter metrics should have "_total" suffix
erlang_vm_memory_bytes_total non-counter metrics should not have "_total" suffix
erlang_vm_memory_processes_bytes_total non-counter metrics should not have "_total" suffix
erlang_vm_memory_system_bytes_total non-counter metrics should not have "_total" suffix
erlang_vm_statistics_run_queues_length_total non-counter metrics should not have "_total" suffix
erlang_vm_atom_count non-histogram and non-summary metrics should not have "_count" suffix
erlang_vm_port_count non-histogram and non-summary metrics should not have "_count" suffix
erlang_vm_process_count non-histogram and non-summary metrics should not have "_count" suffix
erlang_vm_statistics_runtime_milliseconds use base unit "seconds" instead of "milliseconds"
erlang_vm_statistics_wallclock_time_milliseconds use base unit "seconds" instead of "milliseconds"
```

Allowing people to adjust their base units if they like is nice, but IMAO the out-of-the-box defaults should be `promtool` compliant.

NelsonVides:
Fixes #135.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,3 +41,42 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true
           verbose: true
+
+  promtool:
+    name: promtool
+    runs-on: 'ubuntu-24.04'
+    steps:
+      - uses: actions/setup-go@v5
+        with:
+          go-version: 'stable'
+      - run: |
+          # Fetch latest version
+          TAG=$(curl -s https://api.github.com/repos/prometheus/prometheus/releases | jq -r '.[] | .tag_name' | head -n 1)
+          echo $TAG
+          # Remove the `v` prefix
+          VERSION=$(echo $TAG | sed 's/v//')
+          echo $VERSION
+          # Check CPU architecture
+          ARCH=$(uname -s | tr A-Z a-z)-$(uname -m | sed 's/x86_64/amd64/')
+          echo $ARCH
+          # Download the release
+          curl -L -O https://github.com/prometheus/prometheus/releases/download/$TAG/prometheus-$VERSION.$ARCH.tar.gz
+          # Extract the file
+          tar -xvzf prometheus-$VERSION.$ARCH.tar.gz
+          # Move promtool
+          mv prometheus-$VERSION.$ARCH/promtool /usr/local/bin
+          # Verify if promtool is installed
+          promtool --version
+          # Delete downloaded files
+          rm -rf prometheus-$VERSION.$ARCH
+          rm -rf prometheus-$VERSION.$ARCH.tar.gz
+      - uses: actions/checkout@v4
+      - uses: erlef/setup-beam@v1
+        with:
+          otp-version: '28'
+          rebar3-version: '3.25'
+      - run: rebar3 compile
+      - run: |
+          set -e
+          set -o pipefail
+          erl -pa _build/default/lib/prometheus/ebin -pa _build/default/lib/ddskerl/ebin -noshell -eval '{ok, _} = application:ensure_all_started(prometheus), ok = io:format(standard_io, "~s~n", [prometheus_text_format:format()]), halt().' | promtool check metrics
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,10 +15,8 @@ jobs:
     strategy:
       matrix:
         otp: ['28', '27', '26']
-        rebar3: ['3.24', '3.25']
+        rebar3: ['3.25']
     runs-on: 'ubuntu-24.04'
-    env:
-      OTPVER: ${{ matrix.otp }}
     steps:
       - uses: actions/checkout@v4
       - uses: erlef/setup-beam@v1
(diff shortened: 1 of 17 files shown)


.github/workflows/main.yml
```diff
@@ -10,37 +10,76 @@ on:
...
-          token: ${{ secrets.CODECOV_TOKEN }}
-          fail_ci_if_error: true
-          verbose: true
+          otp-version: '28'
+          rebar3-version: '3.25'
+      - run: |
+          set -e
+          set -o pipefail
+          rebar3 shell --eval 'P = "${{runner.temp}}/prom_format" io:format("~s", [P]), file:write_file(P, prometheus_text_format:format()), halt().'
+          cat ${{runner.temp}}/prom_format | promtool check metrics
```
onno-vos-dev: We could test that it currently fails right? And then when we fix it, the pipeline will fail and we simply flip this check 👍

NelsonVides: Yeah, that's what I'm trying to find out how to do, for some reason GHA keeps saying the file does not exist :D

onno-vos-dev: Try (not sure if indentation is correct)
```
- run: |
          set -e
          set -o pipefail
          rebar3 shell --eval 'P = "prom_format" io:format("~s", [P]), file:write_file(P, prometheus_text_format:format()), halt().'
          cat prom_format | promtool check metrics
  working_directory: ${{runner.temp}}
```

NelsonVides: Nope, not working 😅

.github/workflows/main.yml
```diff
@@ -10,37 +10,85 @@ on:
...
-        if: ${{ matrix.otp == '28' }}
-        with:
-          files: _build/test/covertool/prometheus.covertool.xml
-          token: ${{ secrets.CODECOV_TOKEN }}
-          fail_ci_if_error: true
-          verbose: true
+      - run: |
+          set -e
+          set -o pipefail
+          rebar3 shell --eval 'P = "prom_format.txt", ok = file:write_file(P, prometheus_text_format:format()), halt().'
```
paulo-ferraz-oliveira: Maybe you can:

- either pass in the CWD and join to the file, 
- or print the cwd, from within `erl`

?

NelsonVides: This works on my computer just as it is, let's try that 🥲

NelsonVides: https://github.com/prometheus-erl/prometheus.erl/actions/runs/16297341248/job/46022421795 didn't print anything 🤦🏽

.github/workflows/main.yml
```diff
@@ -26,14 +26,13 @@ jobs:
           otp-version: ${{matrix.otp}}
           rebar3-version: ${{matrix.rebar3}}
       - run: rebar3 compile
```
paulo-ferraz-oliveira, line 26: Compile is basically always optional in `rebar3` but you can leave it if you prefer explicitness.

.github/workflows/main.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         otp: ['28', '27', '26']
-        rebar3: ['3.24', '3.25']
+        rebar3: ['3.25']
     runs-on: 'ubuntu-24.04'
     env:
       OTPVER: ${{ matrix.otp }}
```
paulo-ferraz-oliveira: Is this used?

.github/workflows/main.yml
```diff
@@ -26,14 +26,13 @@ jobs:
...
       - run: rebar3 compile
+      - run: rebar3 fmt --check
+        if: ${{ matrix.otp == '28' }}
       - run: rebar3 lint
       - run: rebar3 xref
       - run: rebar3 dialyzer
       - run: rebar3 eunit
-      - run: rebar3 fmt --check
-        if: ${{ matrix.otp == '28' }}
       - run: rebar3 ex_doc
```
paulo-ferraz-oliveira, line 33: Since you're moving stuff, I'd do this last, too.

src/collectors/prometheus_collectors_compat.erl
paulo-ferraz-oliveira, line 1: I like this approach.

paulo-ferraz-oliveira:
It'd be nice (if you remember) if the Release Notes would mentions what needs to be changed, e.g., with a table "From > To" so it's easy to follow.

paulo-ferraz-oliveira (dismissed):

.github/workflows/main.yml
NelsonVides: Nope, probably an old artefact

.github/workflows/main.yml
NelsonVides: Which one you mean, ex_doc? Why so? 😄

NelsonVides:
> It'd be nice (if you remember) if the Release Notes would mentions what needs to be changed, e.g., with a table "From > To" so it's easy to follow.

Hmm... this repo has no `CHANGELOG`, so dunno, probably just the release notes in github will be enough? That one will surely do :)

mikpe (approved):

PR #194, opened 2025-11-03
Add `prometheus_text_format:format_into/3`

the-mikedavis:
This can be used to stream the formatted output, avoiding the buffering of `ram_file` in `format/1`. For example using this library with Cowboy, you could use `cowboy_req:stream_body/3` with the iodata passed to the formatting callback in order to avoid buffering.

Also see https://github.com/rabbitmq/rabbitmq-server/pull/14885, https://github.com/rabbitmq/rabbitmq-server/discussions/14865

---

WIP as I'd like to gather some metrics on memory usage. Also this should probably be added to `prometheus_protobuf_format` for consistency.

### DIFF
--- a/src/formats/prometheus_text_format.erl
+++ b/src/formats/prometheus_text_format.erl
@@ -26,10 +26,10 @@ http_request_duration_milliseconds_sum{method=\"post\"} 4350
 ```
 """).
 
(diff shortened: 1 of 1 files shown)


the-mikedavis:
I folded this into #196 since the changes overlap

PR #196, opened 2025-11-05
Reduce garbage generation in `prometheus_text_format`

the-mikedavis:
This is a somewhat small set of patches to `prometheus_text_format` which aim to reduce garbage creation during registry formatting. Reducing garbage creation drives down the cost to the VM of scraping large registries - both in terms of peak memory allocation and also the work that the garbage collector must do.

With these changes I see reduction in allocation reported by `tprof` in a stress test of one of RabbitMQ's most expensive registries. In a test against single-instance RabbitMQ brokers on EC2 instances this saves a noticeable amount of peak memory and reduces CPU utilization significantly.

<details><summary>tprof testing instructions</summary>

1. Clone `https://github.com/rabbitmq/rabbitmq-server`
2. `cd rabbitmq-server`
3. `make deps`
4. `make run-broker`
5. In another terminal in the `rabbitmq-server` repo, `sbin/rabbitmqctl import_definitions path/to/100k-classic-queues.json` pointing to [this definitions file](https://github.com/rabbitmq/sample-configs/blob/main/queues/100k-classic-queues.json).
6. In the shell from the `make run-broker` terminal, start `tprof` tracing for new processes: `tprof:start(#{type => call_memory}), tprof:enable_trace(new), tprof:set_pattern('_', '_', '_').`
7. In another terminal scrape the expensive endpoint: `curl -v localhost:15692/metrics/per-object --output /dev/null`
8. When that's done, collect and format the sample: `tprof:format(tprof:inspect(tprof:collect())).`

To test this change, <kbd>Ctrl</kbd><kbd>c</kbd> twice out of `make broker`, `cd deps/prometheus` and check out this branch. Then `rm -rf ebin` in that directory, `cd ../../` and repeat steps 4, 6, 7 and 8 again (skipping definitions import).

---

</details>

<details><summary>Registry collection tprof measurement before this change...</summary>

```
****** Process <0.301089.0>  --  100.00% of total *** 
FUNCTION                                                                                   CALLS      WORDS    PER CALL  [    %]
... removed everything less than 1% ...
prometheus_text_format:render_labels/1                                                   2308195    1944642        0.84  [ 1.01]
erlang:atom_to_binary/2                                                                   651584    2375647        3.65  [ 1.23]
prometheus_rabbitmq_core_metrics_collector:'-emit_queue_info/3-fun-0-'/3                  100000    2500000       25.00  [ 1.29]
prometheus_model_helpers:counter_metric/2                                                 301325    3615900       12.00  [ 1.87]
prometheus_text_format:'-render_labels/1-fun-0-'/2                                        321434    4178642       13.00  [ 2.16]
prometheus_rabbitmq_core_metrics_collector:'-collect_metrics/2-lc$^1/1-0-'/2             2300145    4400076        1.91  [ 2.28]
prometheus_model_helpers:'-metrics_from_tuples/2-lc$^0/1-0-'/2                           2308456    4616300        2.00  [ 2.39]
lists:'-filter/2-lc$^0/1-0-'/2                                                           2408461    4816304        2.00  [ 2.49]
erlang:integer_to_binary/1                                                               2206892    6620701        3.00  [ 3.43]
prometheus_rabbitmq_core_metrics_collector:label/1                                       2200038   11000022        5.00  [ 5.69]
prometheus_rabbitmq_core_metrics_collector:'-collect_metrics/2-lc$^0/1-1-'/2             2300145   11500190        5.00  [ 5.95]
prometheus_text_format:'-emit_mf_metrics/2-fun-0-'/3                                     2308150   11541419        5.00  [ 5.97]
prometheus_model_helpers:gauge_metric/2                                                  2006812   24081744       12.00  [12.47]
prometheus_text_format:has_special_char/1                                               23475329   24147190        1.03  [12.50]
prometheus_text_format:render_series/3                                                   2308200   32511401       14.09  [16.83]
ets:match_object/2                                                                            19   38406095  2021373.42  [19.88]
                                                                                                  193184463              [100.0]
```

---

</details>

<details><summary>Registry collection tprof measurement after this change...</summary>

```
****** Process <0.401000.0>  --  99.99% of total *** 
FUNCTION                                                                                  CALLS      WORDS    PER CALL  [    %]
... removed everything less than 1% ...
prometheus_model_helpers:label_pair/1                                                    429393    1717572        4.00  [ 1.16]
prometheus_text_format:render_labels/1                                                  2308195    1944642        0.84  [ 1.32]
erlang:atom_to_binary/2                                                                  651584    2375647        3.65  [ 1.61]
prometheus_rabbitmq_core_metrics_collector:'-emit_queue_info/3-fun-0-'/3                 100000    2500000       25.00  [ 1.69]
prometheus_model_helpers:counter_metric/2                                                301325    3615900       12.00  [ 2.45]
prometheus_text_format:'-render_labels/1-fun-0-'/2                                       321434    4178642       13.00  [ 2.83]
prometheus_rabbitmq_core_metrics_collector:'-collect_metrics/2-lc$^1/1-0-'/2            2300145    4400076        1.91  [ 2.98]
prometheus_model_helpers:'-metrics_from_tuples/2-lc$^0/1-0-'/2                          2308456    4616300        2.00  [ 3.13]
lists:'-filter/2-lc$^0/1-0-'/2                                                          2408461    4816304        2.00  [ 3.26]
erlang:integer_to_binary/1                                                              2206892    6620705        3.00  [ 4.49]
prometheus_rabbitmq_core_metrics_collector:label/1                                      2200038   11000022        5.00  [ 7.45]
prometheus_rabbitmq_core_metrics_collector:'-collect_metrics/2-lc$^0/1-1-'/2            2300145   11500190        5.00  [ 7.79]
prometheus_text_format:render_series/4                                                  2308200   11541000        5.00  [ 7.82]
prometheus_text_format:render_value/2                                                   2308200   11543618        5.00  [ 7.82]
prometheus_model_helpers:gauge_metric/2                                                 2006812   24081744       12.00  [16.32]
ets:match_object/2                                                                           19   38406095  2021373.42  [26.02]
                                                                                                 147597866              [100.0]
```

---

</details>

So with this change, the Cowboy request process in charge of this endpoint allocates `147_597_866` words instead of `193_184_463`, a reduction of `45_586_597` words or 23.6%.

<details><summary>Stress-testing on EC2...</summary>

On EC2 I have two `m7g.xlarge` instances running RabbitMQ: `galactica` which carries this change and `kestrel` which uses `prometheus` at v5.1.1 (latest version RabbitMQ has adopted). A third instance `curl`s these instances at an interval of two seconds with this script:

```sh
#! /usr/bin/env bash

N=600
SLEEP=2
for i in $(seq 1 $N)
do
  echo "Sleeping ${SLEEP}s... ($i / $N)"
  sleep $SLEEP
  echo "Ask for metrics from $1... ($i / $N)"
  curl -s "http://$1:15692/metrics/per-object" --output /dev/null &
done

wait
```

This asynchronously fires off a scrape request every two seconds for twenty minutes. The third node runs this script against both `galactica` and `kestrel` at the same time. The third node also scrapes these nodes' `node_exporter` metrics and RabbitMQ prometheus endpoint for Erlang allocator metrics.

##### Instance-wide memory usage

grafana-kestrel-mem

##### Instance-wide CPU usage

grafana-kestrel-cpu

##### Erlang allocators

grafana-kestrel-erlang-alloc

##### Instance-wide memory usage

grafana-galactica-mem

##### Instance-wide CPU usage

grafana-galactica-cpu

##### Erlang allocators

grafana-galactica-erlang-alloc

---

</details>

We can see `kestrel` (baseline) pinned at around 95% CPU usage consistently, hovering at around 9-10 GB instance-wide memory usage and the VM aware of 3.5-4.5 GB of usage. And `galactica` (this branch) sitting at 50% CPU usage, around 7.5-8.5 GB instance-wide memory and the VM tracking around 2-3 GB of memory.

While the peak memory usage is reduced nicely, the main benefit is the CPU is loaded much less than before - I assume from performing less garbage collection.

### DIFF
--- a/src/formats/prometheus_text_format.erl
+++ b/src/formats/prometheus_text_format.erl
@@ -29,13 +29,13 @@ http_request_duration_milliseconds_sum{method=\"post\"} 4350
 -export([content_type/0, format/0, format/1, render_labels/1, escape_label_value/1]).
 
 -ifdef(TEST).
--export([escape_metric_help/1, emit_mf_prologue/2, emit_mf_metrics/2]).
+-export([escape_metric_help/1]).
 -endif.
 
 -include("prometheus_model.hrl").
 
 -behaviour(prometheus_format).
--compile({inline, [add_brackets/1, render_label_pair/1]}).
+-compile({inline, [render_label_pair/1]}).
 
 ?DOC("""
 Returns content type of the latest \[text format](https://bit.ly/2cxSuJP).
@@ -85,50 +85,40 @@ escape_label_value(Value) ->
     erlang:error({invalid_value, Value}).
 
 registry_collect_callback(Fd, Registry, Collector) ->
-    Callback = fun(MF) ->
-        emit_mf_prologue(Fd, MF),
-        emit_mf_metrics(Fd, MF)
+    Callback = fun(#'MetricFamily'{name = Name0, help = Help, type = Type, metric = Metrics}) ->
+        %% eagerly convert the name to a binary so we can copy more efficiently
+        %% in `render_metrics/3`
+        Name = iolist_to_binary(Name0),
+        Prologue = <<
+            "# TYPE ",
+            Name/binary,
+            " ",
+            (string_type(Type))/binary,
+            "\n# HELP ",
+            Name/binary,
+            " ",
+            (escape_metric_help(Help))/binary,
+            "\n"
+        >>,
+        %% file:write/2 is an expensive operation, as it goes through a port driver.
+        %% Instead a large chunk of bytes is being collected here, in a
+        %% way that triggers binary append optimization in ERTS.
+        file:write(Fd, render_metrics(Name, Metrics, Prologue))
     end,
     prometheus_collector:collect_mf(Registry, Collector, Callback).
 
-?DOC(false).
--spec emit_mf_prologue(Fd :: file:fd(), prometheus_model:'MetricFamily'()) -> ok.
-emit_mf_prologue(Fd, #'MetricFamily'{name = Name, help = Help, type = Type}) ->
-    Bytes = [
-        "# TYPE ",
-        Name,
-        " ",
-        string_type(Type),
-        "\n# HELP ",
-        Name,
-        " ",
-        escape_metric_help(Help),
-        "\n"
-    ],
-    file:write(Fd, Bytes).
+render_metrics(Bytes, _Name, []) ->
+    Bytes;
+render_metrics(Bytes, Name, [Metric | Rest]) ->
+    render_metrics(render_metric(Bytes, Name, Metric), Name, Rest).
 
-?DOC(false).
--spec emit_mf_metrics(file:fd(), prometheus_model:'MetricFamily'()) -> ok | {error, term()}.
-emit_mf_metrics(Fd, #'MetricFamily'{name = Name, metric = Metrics}) ->
-    %% file:write/2 is an expensive operation, as it goes through a port driver.
-    %% Instead a large chunk of bytes is being collected here, in a
-    %% way that triggers binary append optimization in ERTS.
-    Bytes = lists:foldl(
-        fun(Metric, Blob) ->
-            <<Blob/binary, (render_metric(Name, Metric))/binary>>
-        end,
-        <<>>,
-        Metrics
-    ),
-    file:write(Fd, Bytes).
-
-render_metric(Name, #'Metric'{label = Labels, counter = #'Counter'{value = Value}}) ->
-    render_series(Name, render_labels(Labels), Value);
-render_metric(Name, #'Metric'{label = Labels, gauge = #'Gauge'{value = Value}}) ->
-    render_series(Name, render_labels(Labels), Value);
-render_metric(Name, #'Metric'{label = Labels, untyped = #'Untyped'{value = Value}}) ->
-    render_series(Name, render_labels(Labels), Value);
-render_metric(Name, #'Metric'{
+render_metric(Bytes0, Name, #'Metric'{label = Labels, counter = #'Counter'{value = Value}}) ->
+    render_series(Bytes0, Name, render_labels(Labels), Value);
+render_metric(Bytes0, Name, #'Metric'{label = Labels, gauge = #'Gauge'{value = Value}}) ->
+    render_series(Bytes0, Name, render_labels(Labels), Value);
+render_metric(Bytes0, Name, #'Metric'{label = Labels, untyped = #'Untyped'{value = Value}}) ->
+    render_series(Bytes0, Name, render_labels(Labels), Value);
+render_metric(Bytes0, Name, #'Metric'{
     label = Labels,
     summary = #'Summary'{
         sample_count = Count,
@@ -137,12 +127,13 @@ render_metric(Name, #'Metric'{
     }
 }) ->
     LString = render_labels(Labels),
-    Bytes1 = render_series([Name, "_count"], LString, Count),
-    Bytes2 = <<Bytes1/binary, (render_series([Name, "_sum"], LString, Sum))/binary>>,
-    Bytes3 = lists:foldl(
+    Bytes1 = render_series(Bytes0, <<Name/binary, "_count">>, LString, Count),
+    Bytes2 = render_series(Bytes1, <<Name/binary, "_sum">>, LString, Sum),
+    lists:foldl(
         fun(#'Quantile'{quantile = QN, value = QV}, Blob) ->
-            Val = render_series(
-                [Name],
+            render_series(
+                Blob,
+                Name,
                 render_labels(
                     [
                         LString,
@@ -153,14 +144,12 @@ render_metric(Name, #'Metric'{
                     ]
                 ),
                 QV
-            ),
-            <<Blob/binary, Val/binary>>
+            )
         end,
         Bytes2,
         Quantiles
-    ),
-    Bytes3;
-render_metric(Name, #'Metric'{
+    );
+render_metric(Bytes0, Name, #'Metric'{
     label = Labels,
     histogram = #'Histogram'{
         sample_count = Count,
@@ -172,33 +161,35 @@ render_metric(Name, #'Metric'{
     LString = render_labels(Labels),
     Bytes1 = lists:foldl(
         fun(Bucket, Blob) ->
-            <<Blob/binary, (emit_histogram_bucket(Name, LString, Bucket))/binary>>
+            emit_histogram_bucket(Blob, Name, LString, Bucket)
         end,
-        <<>>,
+        Bytes0,
         Buckets
     ),
-    Bytes2 = <<Bytes1/binary, (render_series([Name, "_count"], LString, Count))/binary>>,
-    Bytes3 = <<Bytes2/binary, (render_series([Name, "_sum"], LString, Sum))/binary>>,
-    Bytes3.
+    Bytes2 = render_series(Bytes1, <<Name/binary, "_count">>, LString, Count),
+    render_series(Bytes2, <<Name/binary, "_sum">>, LString, Sum).
 
-emit_histogram_bucket(Name, LString, #'Bucket'{cumulative_count = BCount, upper_bound = BBound}) ->
+emit_histogram_bucket(Bytes0, Name, LString, #'Bucket'{
+    cumulative_count = BCount, upper_bound = BBound
+}) ->
     BLValue = bound_to_label_value(BBound),
     render_series(
-        [Name, "_bucket"],
+        Bytes0,
+        <<Name/binary, "_bucket">>,
         render_labels([LString, #'LabelPair'{name = "le", value = BLValue}]),
         BCount
     ).
 
 string_type('COUNTER') ->
-    "counter";
+    <<"counter">>;
 string_type('GAUGE') ->
-    "gauge";
+    <<"gauge">>;
 string_type('SUMMARY') ->
-    "summary";
+    <<"summary">>;
 string_type('HISTOGRAM') ->
-    "histogram";
+    <<"histogram">>;
 string_type('UNTYPED') ->
-    "untyped".
+    <<"untyped">>.
 
 %% binary() in spec means 0 or more already rendered labels (name,
 %% escaped value), joined with "," in between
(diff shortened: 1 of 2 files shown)


NelsonVides:
This looks amazing! I see it is still marked as draft so I guess no rush, and I'm away from the computer for a few days so only having a look at this on my phone now. Nevertheless, I'd love to see this ready and merged, thank you so much for the work 😃

the-mikedavis:
Yep no real rush on this! Looks like I have some work to do to make the CI happy anyways

lhoguin:
I've approved the workflow run.

NelsonVides (commented):
I'm still away from a proper computer so still quickly looking at this from my phone. Thanks a lot for having it pass CI!

I have a question, for good-looking code, the process dictionary's a bit ugly. Is there a way to refactor those usages to pass something more functional or does it have a performance impact that way?

Also for the compiled regex, perhaps storing it in a persistent term that is created at startup could help that perform even faster? It would literally be compiled only once through the entire VM lifetime and never require any GC.

If you don't see an easy way to improve that then it's probably fine to merge this way and later when I'm back to a full computer I'll try to refactor and tag you in a potential PR 🤔

the-mikedavis:
Yeah the process dictionary part here is really icky. I thought in my earlier testing that I saw `persistent_term:get/2` allocating but, trying again, looks like I was wrong. So the `binary:match/2` pattern we should definitely move into a persistent_term.

For the `erase/1`+`put/2` dance in `format_into/3` we can't be more functional because the `Collector:collect_mf/2` callback returns `ok`, so we don't have a nice way to accumulate values. To fix that we'd need a really big breaking change :/

src/formats/prometheus_text_format.erl
```diff
@@ -84,51 +114,18 @@ escape_label_value(LValue) when is_list(LValue) ->
...
-            <<Blob/binary, (render_metric(Name, Metric))/binary>>
-        end,
-        <<>>,
-        Metrics
-    ),
-    file:write(Fd, Bytes).
+render_metrics(_Name, [], Bytes) ->
+    Bytes;
+render_metrics(Name, [Metric | Rest], Bytes) ->
+    render_metrics(Name, Rest, render_metric(Bytes, Name, Metric)).
```
NelsonVides: I like how this becomes tail-recursive. Note nevertheless, that if we want to be really mad about performance, the shuffling in the params ordering here is not ideal. The BEAM assigns each function input to X registers, so before calling `render_metric(Bytes, Name, Metric)`, it would have to shuffle X0 to X1, and X2 to X0. The JIT will turn this into CPU registers so it is insanely fast, but it will not reorder them into a faster version because it would mess up with tracing. So instead we could be calling here `render_metric(Name, Metric, Bytes)` and the only argument arrangement would be by extracting `Metric` into X1. Likely to be in the order of 3 CPU cycles the whole thing, but `gen_server` code takes great care to do this right so, why not 😄

src/formats/prometheus_text_format.erl
```diff
@@ -57,16 +57,46 @@ format() ->
...
+-spec format_into(Registry :: prometheus_registry:registry(), term(), fun(
+    (binary(), term()) -> term()
+)) -> term().
+format_into(Registry, State, Fmt) ->
+    State1 = lists:foldl(
+        fun(Collector, Acc) ->
+            put(?MODULE, Acc),
+            Callback = fun(
+                #'MetricFamily'{name = Name0, help = Help, type = Type, metric = Metrics}
+            ) ->
```
NelsonVides: I tend to do tracing often for debugging runtime systems and anonymous functions are then one of the most unreadable parts of the whole experience. They will print as something like `format_info-1-3942/2` `format_info-1-3294/2`, which is quite unreadable to figure out which is which and what does it actually do. Unfortunately here the anonymous functions are capturing environment so we can't just create a plain "`format_info_collect_mf_callback/2`", but we could have a "`format_info_collect_mf_callback/3`" that actually takes the variables we're going to capture and returns the anonymous function as-is. Then tracing will show "`format_info_collect_mf_callback-1-4324/2`" which ain't perfect but at least has a unique and expressive enough name. Same idea for the other anonymous function in this part.

src/formats/prometheus_text_format.erl
```diff
@@ -57,16 +57,46 @@ format() ->
...
-    Callback = fun(_, Collector) ->
-        registry_collect_callback(Fd, Registry, Collector)
-    end,
-    prometheus_registry:collect(Registry, Callback),
-    file:write(Fd, "\n"),
-    {ok, Size} = ram_file:get_size(Fd),
-    {ok, Str} = file:pread(Fd, 0, Size),
-    ok = file:close(Fd),
-    Str.
+    format_into(Registry, <<>>, fun(Data, Acc) -> <<Acc/binary, Data/binary>> end).
```
NelsonVides: As in the other comment about anonymous functions but this one is easy:
```suggestion
    format_into(Registry, <<>>, fun format_into_acc/2).

format_into_acc(Data, Acc) ->
    <<Acc/binary, Data/binary>>.
```

Provided any better naming you could think of. Also, why not swapping the param order? Isn't it more readable 

```erl
format_into_acc(Acc, Data) ->
    <<Acc/binary, Data/binary>>.
```

And likewise somewhere below

```erl
    Fmt(State1, <<"\n">>).
```

That way it is more intuitive that what we're probably doing is appending.

NelsonVides (commented):
There's so many checks against whether something is a binary and if it is not then `iolist_to_binary`, that I really wish everything had been just binaries to begin with. But that's a breaking change for another day 😄

Anyway, I have a couple of comments regarding a bit more crazy performance optimisations and making tracing more readable. It's not really important and if it is too annoying I could just shuffle this code myself in a couple of weeks. But maybe you like the idea or you didn't know those tricks and want to do it yourself so just sharing. WDYT? 🙂

the-mikedavis:
All of that sounds good to me - I applied all of the suggestions. For the `render_metrics/3` one I tweaked it slightly so that the binary is always the first arg

src/formats/prometheus_text_format.erl
```diff
@@ -233,29 +239,17 @@ render_label_pair(B) when is_binary(B) ->
...
-        " ",
-        (iolist_to_binary(io_lib:format("~p", [Value])))/binary,
-        "\n"
-    >>.
+render_value(Bytes, undefined) ->
+    <<Bytes/binary, "NaN\n">>;
+render_value(Bytes, Value) when is_integer(Value) ->
+    <<Bytes/binary, (integer_to_binary(Value))/binary, "\n">>;
+render_value(Bytes, Value) ->
+    <<Bytes/binary, (iolist_to_binary(io_lib:format("~p", [Value])))/binary, "\n">>.
```
NelsonVides, line 255: Is there any scenario why this wouldn't be a float? IIRC all values will either be undefined, an integer, or a float. If that assumption is correct, and considering the prometheus protocol actually loves floats, `float_to_binary/1-2` would be much faster and allocate no intermediate garbage here 🤔

NelsonVides (commented):
Loving it. Got one more question :D

src/formats/prometheus_text_format.erl
```diff
@@ -172,33 +175,36 @@ render_metric(Name, #'Metric'{
...
+            emit_histogram_bucket(Blob, Name, LString, Bucket)
         end,
-        <<>>,
+        Bytes0,
         Buckets
     ),
-    Bytes2 = <<Bytes1/binary, (render_series([Name, "_count"], LString, Count))/binary>>,
-    Bytes3 = <<Bytes2/binary, (render_series([Name, "_sum"], LString, Sum))/binary>>,
+    Bytes2 = render_series(Bytes1, <<Name/binary, "_count">>, LString, Count),
+    Bytes3 = render_series(Bytes2, <<Name/binary, "_sum">>, LString, Sum),
```
NelsonVides: Also this could probably just return the result without binding the `Bytes3` variable 🤔

src/formats/prometheus_text_format.erl
```diff
@@ -137,12 +140,13 @@ render_metric(Name, #'Metric'{
     }
 }) ->
     LString = render_labels(Labels),
-    Bytes1 = render_series([Name, "_count"], LString, Count),
-    Bytes2 = <<Bytes1/binary, (render_series([Name, "_sum"], LString, Sum))/binary>>,
+    Bytes1 = render_series(Bytes0, <<Name/binary, "_count">>, LString, Count),
+    Bytes2 = render_series(Bytes1, <<Name/binary, "_sum">>, LString, Sum),
```
NelsonVides, line 144: Likewise at the end of the fold below, no need to bind Bytes3.

NelsonVides (commented):
Just dirty stuff, this was there before your code changes anyway, it just now shows on the diff.

src/formats/prometheus_text_format.erl
the-mikedavis: Looks like `'Gauge'`, `'Counter'`, `'Untyped'` etc allow for a few other values (looking at `prometheus_model.hrl`). We could add branches for those too though - avoiding `io_lib:format/2` would most likely be a nice improvement. Or even just adding a branch for floats.

NelsonVides: Well yeah, floats will very likely be a hot path, other values would rather not happen much in reality.

Checking the docs, these seem like the possible values: https://prometheus.io/docs/specs/om/open_metrics_spec_2_0/#values, where numbers are described at: https://prometheus.io/docs/specs/om/open_metrics_spec_2_0/#numbers so whatever is the right way to print floats using https://www.erlang.org/doc/apps/erts/erlang.html#float_to_binary/2 will surely help performance here.

the-mikedavis: Looks like this is how it's currently done: https://github.com/prometheus-erl/prometheus.erl/blob/a4f14320781787f838b2ddf8f886cd02c279bee7/src/formats/prometheus_text_format.erl#L276

That should probably also use binaries instead.

the-mikedavis: To match `io_lib:format/2` I added a branch for floats that uses `float_to_binary(Value, [short])`, and a TODO comment about it. Maybe we should update that `bound_to_label_value/1` function to use `[short]` instead. But I'd like to avoid any possibly breaking changes with this PR on the other hand.

We still make some garbage for this with `float_to_binary/2` but avoiding all of that allocation in `io_lib:format/2` is a nice win:

```
1> tprof:profile(erlang, float_to_binary, [1520879607.789], #{type => call_memory}).

****** Process <0.87.0>  --  100.00% of total *** 
FUNCTION                  CALLS  WORDS  PER CALL  [     %]
erlang:float_to_binary/1      1      6      6.00  [100.00]
                                     6            [ 100.0]
ok
2> tprof:profile(io_lib, format, ["~p", [1520879607.789]], #{type => call_memory}).

****** Process <0.93.0>  --  100.00% of total *** 
FUNCTION                       CALLS  WORDS  PER CALL  [    %]
io_lib_format:collect/2            2      2      1.00  [ 1.38]
io_lib_format:build_small/1        2      2      1.00  [ 1.38]
io_lib_format:build_limited/5      2      2      1.00  [ 1.38]
io_lib_format:modifiers/3          1      4      4.00  [ 2.76]
io_lib_format:precision/2          1      4      4.00  [ 2.76]
io_lib_format:field_value/2        1      4      4.00  [ 2.76]
io_lib_format:pad_char/2           1      4      4.00  [ 2.76]
io_lib_format:field_width/3        1      5      5.00  [ 3.45]
io_lib_format:collect_cc/2         1      7      7.00  [ 4.83]
io_lib_format:count_small/2        2     12      6.00  [ 8.28]
erlang:float_to_list/2             1     28     28.00  [19.31]
io_lib_format:print/11             3     35     11.67  [24.14]
io_lib_format:collect_cseq/2       1     36     36.00  [24.83]
                                        145            [100.0]
ok
```

NelsonVides: Right, that's all wonderful findings. Thank you so much for this change!

NelsonVides (approved):
A blast of a change! Thank you so much for all the effort and also, for keeping such a tidy git history :)

NelsonVides:
@the-mikedavis gonna get a release to hex done over the weekend 👍🏽

the-mikedavis:
Sweet, thanks @NelsonVides!

NelsonVides:
Published https://hex.pm/packages/prometheus/6.1.0 🎉

src/formats/prometheus_text_format.erl
```diff
@@ -57,16 +57,52 @@ format() ->
...
+            " ",
+            (string_type(Type))/binary,
+            "\n# HELP ",
+            Name/binary,
+            " ",
+            (escape_metric_help(Help))/binary,
+            "\n"
+        >>,
+        Bin = render_metrics(Prologue, Name, Metrics),
+        put(?MODULE, Fmt(Bin, erase(?MODULE)))
```
the-mikedavis, line 104: Bah, I made a mistake here with the order of the arguments. The function should take the state as the first argument and then the new data as the second argument. It doesn't end up making a difference for `format/1` because it just changes the order that the metrics families are formatted in - it's just concatenating the wrong way. But using `format_into/3` with a custom formatting function doesn't work properly. I'll send a follow-up PR (edit: #197)

PR #197, opened 2025-11-22
prometheus_text_format: Fix argument order for `format_into/3` fun

the-mikedavis:
The formatting function should take the state as the first argument and new data as the second. See `format/1` which passes in the trailing linefeed as the second argument. This doesn't make much difference when calling `format/1` because the swapped arguments only caused the metrics to be concatenated in reverse. But this fix is important for using `format_into/3` correctly, for example with `cowboy_req:stream_body/3`.

### DIFF
--- a/src/formats/prometheus_text_format.erl
+++ b/src/formats/prometheus_text_format.erl
@@ -101,7 +101,7 @@ format_into_create_mf_callback_fn(Fmt) ->
             "\n"
         >>,
         Bin = render_metrics(Prologue, Name, Metrics),
-        put(?MODULE, Fmt(Bin, erase(?MODULE)))
+        put(?MODULE, Fmt(erase(?MODULE), Bin))
     end.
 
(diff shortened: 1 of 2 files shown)


NelsonVides (approved):
Great to have a regression test for it too!

NelsonVides:
https://hex.pm/packages/prometheus/6.1.1 😁

the-mikedavis:
Thanks for the super quick turnaround on this! I'm re-running my tests in RabbitMQ with 6.1.1. I can confirm that it works now 😅. And so far the comparison is still looking very good - I think the per-object metrics RabbitMQ end up using quite a few floats so the fast-path we added in #196 probably cuts down on temporary allocations by quite a bit.

Edit: yep with these changes and `cowboy_req:stream_body/3` this is looking great! https://github.com/rabbitmq/rabbitmq-server/pull/14885#issuecomment-3567078883

PR #202, opened 2026-04-14
Optimize text format scrape hot path

NelsonVides:
Profiling (tprof) of a RabbitMQ workload with ~2.3M metric series per scrape revealed that a significant portion of scrape time is spent on repeated conversions and intermediate allocations that can be avoided. This spawned from https://github.com/rabbitmq/rabbitmq-server/pull/14885, which spawned itself into https://github.com/prometheus-erl/prometheus.erl/pull/196. I have more changes in mind but they'll be breaking so for now this PR proposes only backwards-compatible stuff:

- Pre-compute metric name/help as binaries at declaration time — `atom_to_binary/iolist_to_binary` were called per metric family on every scrape; now done once at `insert_mf` time and stored in ETS. Includes backward-compat normalization for old ETS entries.
- Inline and fast-path metric record creation — `gauge_metric/2`, `counter_metric/2`, and `label_pair/1` are inlined (these are ~2M calls/scrape at 12 words each). A fast path in `create_mf/4` skips `metrics_from_tuples` dispatch when metrics are already `#'Metric'{}` records. Numeric clauses added to `ensure_binary_or_string/1` to avoid `io_lib:format` fallback.
- Merge filter and map into single list traversal — `metrics_from_tuples/2` previously ran `lists:filter/2` then a comprehension (two passes over ~2.4M elements, `lists:filter/2` alone at 3.26% of scrape time). Now a single comprehension with a guard.
- Inline `render_series/4` and `render_value/2` — called ~2.3M times each per scrape (~15.6% combined), inlining eliminates call overhead and enables cross-boundary binary append optimization.

### DIFF
--- a/src/model/prometheus_model_helpers.erl
+++ b/src/model/prometheus_model_helpers.erl
@@ -45,7 +45,6 @@ Probably will be used with `m:prometheus_collector`.
 
 -ifdef(TEST).
 -export([
-    filter_undefined_metrics/1,
     ensure_mf_type/1,
     ensure_binary_or_string/1
 ]).
@@ -106,12 +105,11 @@ create_mf(Name, Help, Type, [First | _] = Metrics0) when
     is_list(Metrics0), is_record(First, 'Metric')
 ->
     %% Fast path: metrics are already #'Metric'() records, skip metrics_from_tuples
-    Metrics = filter_undefined_metrics(Metrics0),
     #'MetricFamily'{
         name = ensure_binary_or_string(Name),
         help = ensure_binary_or_string(Help),
         type = ensure_mf_type(Type),
-        metric = Metrics
+        metric = [M || M <- Metrics0, M =/= undefined]
     };
 create_mf(Name, Help, Type, Metrics0) ->
     Metrics = metrics_from_tuples(Type, Metrics0),
@@ -389,10 +387,7 @@ histogram_bucket({Bound, Count}) ->
     }.
 
 metrics_from_tuples(Type, Metrics) ->
-    [
-        metric_from_tuple(Type, Metric)
-     || Metric <- filter_undefined_metrics(ensure_list(Metrics))
-    ].
+    [metric_from_tuple(Type, M) || M <- ensure_list(Metrics), M =/= undefined].
 
 metric_from_tuple(_, Metric) when is_record(Metric, 'Metric') ->
     Metric;
@@ -412,14 +407,6 @@ metric_from_tuple(untyped, Metric) ->
 -spec ensure_list(Val :: term()) -> list().
 ensure_list(Val) when is_list(Val) -> Val;
 ensure_list(Val) -> [Val].
-
-?DOC(false).
--spec filter_undefined_metrics([undefined | T]) -> [T].
-filter_undefined_metrics(Metrics) ->
-    lists:filter(fun not_undefined/1, Metrics).
-
-not_undefined(undefined) -> false;
-not_undefined(_) -> true.
 
 ?DOC(false).
 -spec ensure_binary_or_string(Val :: term()) -> binary() | string().
--- a/src/model/prometheus_model_helpers.erl
+++ b/src/model/prometheus_model_helpers.erl
@@ -53,6 +53,8 @@ Probably will be used with `m:prometheus_collector`.
 
 -include("prometheus_model.hrl").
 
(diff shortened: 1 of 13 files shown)


NelsonVides:
@the-mikedavis WDYT about the above ☝🏽😄

src/model/prometheus_model_helpers.erl
```diff
@@ -340,9 +351,9 @@ fail with an error.
 label_pairs(B) when is_binary(B) ->
     B;
 label_pairs(Labels) when is_list(Labels) ->
-    lists:map(fun label_pair/1, Labels);
+    [label_pair(L) || L <- Labels];
 label_pairs(Labels) when is_map(Labels) ->
-    lists:map(fun label_pair/1, maps:to_list(Labels)).
+    [label_pair(L) || L <- maps:to_list(Labels)].
```
the-mikedavis: We could avoid the extra allocations from conversion here if we don't mind depending on OTP26+ for https://www.erlang.org/eeps/eep-0058


```suggestion
    [#'LabelPair'{
         name = ensure_binary_or_string(Name),
         value = ensure_binary_or_string(Value)
     }
     ||
     Name := Value <- Labels
    ].
```

the-mikedavis (commented):
Looks promising! I will rerun my tests with RabbitMQ

lukebakken (commented):
Hello! I did a drive-by review with the 🧞 

---

### `collect_metrics/2` first argument is now always a binary

`create_mf/5` calls `Collector:collect_metrics(Name, CollectorData)` passing whatever was given as its first argument. After this PR, all in-tree metric modules call `create_mf(NameBin, HelpBin, ..., ?MODULE, Data)` with a pre-computed binary, so `collect_metrics/2` now always receives a binary as its first argument.

The in-tree implementations all ignore it (`_NameBin`) and recover the original name from the data tuple, so nothing breaks here. But the documented pattern in `prometheus_collector` shows:

```erlang
collect_metrics(erlang_vm_bytes_total, Memory) ->
    ...

create_gauge(Name, Help, Data) ->
    prometheus_model_helpers:create_mf(Name, Help, gauge, ?MODULE, Data).
```

An external collector following this pattern - with atom-matched `collect_metrics/2` clauses and `create_mf/5` - would get a function clause error at scrape time if `Name` is now a binary. The `collect_metrics/2` callback spec says `Name :: prometheus_metric:name()` which permits binaries, so it is not technically a spec violation, but the behaviour change is silent and the documented example does not prepare implementers for it.

Is this an intentional interface change? If so, it may be worth a note in the changelog and an update to the `prometheus_collector` doc example to show atom-matched clauses are no longer safe.

### Backward-compat normalization path has no test coverage

`normalize_mf_row/1` has two clauses: the new 3-tuple fast path and a backward-compat clause for old ETS entries that still hold the 2-tuple `{Labels, Help}` shape. Codecov confirms the backward-compat clause is not exercised by the test suite.

This clause is only reachable during a hot upgrade from a pre-PR node. A test that inserts a raw 2-tuple ETS entry and calls `metrics/2` would cover it.

src/model/prometheus_model_helpers.erl
NelsonVides: I think we already require the latest 3 versions, so that's 26-27-28 as of today, and soon it'll be 27-28-29 🔥

Applying this change 👍🏽

lukebakken (approved):
🚀 :shipit:

NelsonVides:
Thanks @lukebakken! Waiting for some RMQ tests then @the-mikedavis 🔥

mkuratczyk:
Thanks a lot for working on this. The changes look good to me, but at the same time, I don't see any meaningful improvements when scraping /metrics/per-object with [100k classic queues](https://github.com/rabbitmq/sample-configs/blob/main/queues/100k-classic-queues.json) imported. I will take another look in a bit, but can you share how you tested it and what your results were?

mkuratczyk (approved):
While I don't see an overall improvement in the time to scrape, flamgraphs confirm less time spent in `metrics_from_tuples` and `lists:filter`. Thanks!

NelsonVides:
I don't have services with that gigantic amount of metrics so I didn't have a measurable time difference other than `tprof` profiles saying it actually took less time and memory overall :)

Thanks for running the tests, good to confirm no regressions at least!

PR #203, opened 2026-04-23
Add API to set default value of labeled metrics

fen-pl:
This PR adds a capability to initialize a metric with labels with a default value for specific label values. This is not done automatically, as label values are not known at the moment of metric creation, so it's up to the application whether these should be set to initial default value or not.

### DIFF
--- a/src/metrics/prometheus_boolean.erl
+++ b/src/metrics/prometheus_boolean.erl
@@ -41,7 +41,7 @@ fuse_event(Fuse, Event) ->
     declare/1,
     deregister/1,
     deregister/2,
-    set_default/2,
+    set_default/3,
     set/2,
     set/3,
     set/4,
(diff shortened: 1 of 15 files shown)


fen-pl:
Ping :)

NelsonVides:
Woops, part of the delay was my holidays, part I didn't notice! Thanks a lot for the contribution, noted as a TODO for next week 💪🏽

NelsonVides:
In the meantime, note CI please 😄 I updated the master branch to run CI, feel free to rebase on top of the latest master and remove the merge commits 👌🏽

src/prometheus_metric.erl
```diff
@@ -85,10 +87,11 @@ as well as handling metric labels and data.
...
 
-?DOC("Sets the default metric function for the module.").
--callback set_default(Registry, Name) -> dynamic() when
-    Registry :: prometheus_registry:registry(),
-    Name :: name().
+?DOC("Sets the default (zero/initial) state for the metric series identified by Registry, Name and LabelValues.").
+-callback set_default(Registry :: prometheus_registry:registry(),
+                      Name :: name(),
+                      LabelValues :: label_values()) ->
+    dynamic().
```
NelsonVides, line 94: An issue with this is that it is a breaking change and according to SemVer I'd have to release v7, when it doesn't need to be. I think that `set_default/2` might continue to exist nevertheless, and by default call `set_default/3` with an empty list of labels. That way this PR only adds a new callback without removing an existing one, therefore not breaking any behaviour.

src/metrics/prometheus_boolean.erl
```diff
@@ -127,10 +127,22 @@ deregister(Registry, Name) ->
...
+* `{unknown_metric, Registry, Name}` error if boolean with name `Name` can't be found in `Registry`.
+* `{invalid_metric_arity, Present, Expected}` error if labels count mismatch.
+""").
+-spec set_default(
+    Registry :: prometheus_registry:registry(),
+    Name :: prometheus_metric:name(),
+    LabelValues :: prometheus_metric:label_values()
+) -> ok.
+set_default(Registry, Name, LabelValues) ->
+    set(Registry, Name, LabelValues, undefined).
```
NelsonVides, line 145: This actually looks like an old hidden bug... all other `set_default` changes use `ets:insert_new/` while only here we will do `ets:insert/` and potentially override already existing values 🫣
Could you try a fix with a regression test for this one? 🙏🏽

test/eunit/metric/prometheus_counter_tests.erl
```diff
@@ -312,3 +313,33 @@ test_collector3(_) ->
...
+    ]),
+    %% Before seeding: labeled series is undefined
+    Undef = prometheus_counter:value(http_requests_total, [get]),
+    %% Seed the labeled series
+    prometheus_counter:set_default(default, http_requests_total, [get]),
+    %% After seeding: labeled series returns 0
+    Seeded = prometheus_counter:value(http_requests_total, [get]),
+    %% Idempotent: second call returns false, series is still 0
+    Idempotent = prometheus_counter:set_default(default, http_requests_total, [get]),
+    AfterIdempotent = prometheus_counter:value(http_requests_total, [get]),
```
NelsonVides, line 331: We should be also calling `set` and then checking that `set_default` would not override an already existing value here

NelsonVides (changes requested):
A few comments, and thanks again for the PR and sorry for the delay :)

PR #204, opened 2026-05-30
fix(prometheus): replace deprecated catch with try/catch

asyncmind0:
Update quantile summary reset logic to avoid the deprecated Erlang `catch Expr` form.

The reset operation now uses an explicit `try ... catch ... end` block, preserving the previous failure-tolerant behavior while removing the compiler warning.


paulo-ferraz-oliveira (approved):

paulo-ferraz-oliveira:
@NelsonVides, this seems Ok to me, but I'll let you decide, since I'm not so active in this repository.

NelsonVides (approved):

NelsonVides:
Woop, yes! Sorry didn't notice that one, absolutely needed!

NelsonVides:
https://hex.pm/packages/prometheus/6.1.3 🎉

